PR #7, opened 2018-02-25
Add second recording to readme.

gkbrk:


jonhoo:
Thanks! `README.md` is actually currently autogenerated from the crate-level docs [in `src/lib.rs`](https://github.com/jonhoo/tsunami/blob/379acb77aa8b71a4c92d91342ce76d8a9572e26e/src/lib.rs#L47-L48), so you'd need to modify those and run [`cargo readme > README.md`](https://github.com/livioribeiro/cargo-readme).

Since you're modifying it anyway, could you also make the links more link-y:
```
You can find the recordings of past sessions below:
 - [Part 1](https://youtu.be/Zdudg5TV9i4)
 - [Part 2](https://youtu.be/66INYb73yXo)
```

I think I'd also like to add a note there about the video descriptions containing links to useful intermediate points, plus a summary of what's covered in each part, but I can do that separately from this change :)

gkbrk:
@jonhoo Should be done now

Issue #3, opened 2018-02-25
Enable user to supply region and credentials provider


jonhoo:
At the moment we hard-code the credentials provider and instance region

https://github.com/jonhoo/tsunami/blob/379acb77aa8b71a4c92d91342ce76d8a9572e26e/src/lib.rs#L249-L250

While that makes it easy to get up and running, some users may need to authenticate through more elaborate means (e.g., through `sts::AssumeRole`) as discussed [here](https://github.com/rusoto/rusoto/blob/master/AWS-CREDENTIALS.md). Similarly, not all users will want to use the US-EAST-1 region for obvious reasons. We should provide a way for the user to specify these options in the builder, perhaps also including the [connector](https://github.com/jonhoo/tsunami/blob/379acb77aa8b71a4c92d91342ce76d8a9572e26e/src/lib.rs#L248) to use.

jonhoo:
cc @Cosaquee

matthewkmayer:
Once https://github.com/rusoto/rusoto/issues/957 is implemented in Rusoto, it should be easier to hand-wave away any hardcoding of regions by using a default region lookup like other AWS SDKs provide. 👍

PR #8, opened 2018-02-27
Adds region selection to TsunamiBuilder

simonrw:
Implement custom region selection for Tsunami

This goes half way to solving #3. It allows the user to select a
different aws region from us-east-1.

The documentation for the method includes a link to the
`rusoto_core::Region` struct, and displays the default value.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -191,7 +191,8 @@ impl TsunamiBuilder {
     /// Set up the machines in a specific EC2
     /// [`Region`](http://rusoto.github.io/rusoto/rusoto_core/region/enum.Region.html).
     ///
-    /// The default region is us-east-1.
+    /// The default region is us-east-1. [Available regions are listed
+    /// here](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-available-regions)
     pub fn in_region(&mut self, region: Region) {
         self.region = region;
     }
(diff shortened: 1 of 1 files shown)


simonrw:
There is the question of: is this the desired behaviour? Should the code default to `us-east-1`, or error when the region is not supplied? It seems quite a fundamental piece of information from the user, that perhaps it makes sense making it a critical error if it's _not_ supplied?

jonhoo:
This looks good! I think it'd also be nice to include a link to https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-available-regions.

As far as your question goes: yes, it's not entirely clear that any one region is a more sensible default than the others. I think the right thing here (and probably in general) is to provide convenience rather than forcing the user to make a choice, even when there is no good default value. One way to make this choice is to think about the downsides of the default being wrong. If the downsides are severe, then it might make sense to not have a default and force a choice (i.e., by erroring if it's not given). In our case though, the downsides are fairly minor (ssh latency will be a bit higher), except in the case where the instances need to talk to some *particular* third party, where the user explicitly selecting a region seems like a reasonable cost.

simonrw:
I've updated the docs, and renamed the method to `set_region` which fits in more with the naming convention of the other methods.

I think you're right with the lack of defaults. My only concern is unexpected cost differences, but I don't think AWS pricing varies that much between regions.

jonhoo:
If we move to a builder pattern more in line with [`C-BUILDER` in the Rust API guidelines](https://rust-lang-nursery.github.io/api-guidelines/type-safety.html#c-builder) then we'll probably pick more ergonomic method names, but I think `set_*` is fine for now. Thanks!

jonhoo:
@mindriot101 see also https://github.com/rusoto/rusoto/issues/957

PR #12, opened 2018-03-02
Do not cancel one-time spot requests

### LINKED ISSUE #11: Don't unnecessarily cancel spot instance requests
The default type of spot instance request is [one-time](https://rusoto.github.io/rusoto/rusoto_ec2/struct.RequestSpotInstancesRequest.html#structfield.type_), which means that you shouldnt need to cancel it once it was created -- it will only be fulfilled once.

freshstrangemusic:
The default behaviour of the spot instance request API is to issue a
one-time request so they do not have to be cancelled afte rthey are
fufilled.

Fixes #11

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -337,10 +337,15 @@ impl TsunamiBuilder {
 
             // TODO: VPC
 
-            let mut req = rusoto_ec2::RequestSpotInstancesRequest::default();
-            req.instance_count = Some(i64::from(number));
-            req.block_duration_minutes = Some(self.max_duration);
-            req.launch_specification = Some(launch);
+            let req = rusoto_ec2::RequestSpotInstancesRequest {
+                instance_count: Some(i64::from(number)),
+                block_duration_minutes: Some(self.max_duration),
+                launch_specification: Some(launch),
+                // one-time spot instances are only fulfilled once and therefore do not need to be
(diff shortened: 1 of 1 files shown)


jonhoo:
Just for clarification, can you also specifically set the spot instance request type to `one-time` in https://github.com/brennie/tsunami/blob/f75ca3591064561ec164d7b50100f38033962177/src/lib.rs#L343, along with a comment saying "so we don't have to explicitly cancel the request"?

jonhoo:
I think specifically you [have to set](https://rusoto.github.io/rusoto/rusoto_ec2/struct.RequestSpotInstancesRequest.html#structfield.type_):
```rust
req.type_ = Some("one-time".to_string());
```
It'd be nice if this used an enum or something instead of a string, but we'll work with what we've got :)

freshstrangemusic:
Done & done.

jonhoo:
I think you accidentally got rid of the part of the commit that removes the spot request cancellation. Your changes now *only* sets `one-time`, but does *not* remove the cancellation request :)

freshstrangemusic:
Oops!

freshstrangemusic:
Not sure how I pulled that one off :P But It's been updated with both changes now.

jonhoo:
Great, thanks!

PR #20, opened 2018-04-06
Refactor run_as into submethods

### LINKED ISSUE #19: Separate run_as into many smaller components
Currently, `run_as` is quite long, and has many relatively independent pieces to it (see also [this comment](https://www.youtube.com/watch?v=1qv1L8t2-W0&lc=UgwDf-qonMNykWUC68V4AaABAg)). Especially on the `async` branch. Splitting that function into many smaller ones would make the code much easier to read and manage! It shouldn't be *too* bad to take most of the closures and split them out into separate private `&self` functions on `TsunamiBuilder`, so we should probably do that!

ViktorHavrysh:
Fixes #19.

Note that 1) I'm pretty new to rust and 2) I haven't run this yet -- need to set up AWS on my home machine...

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -161,6 +161,17 @@ impl MachineSetup {
     }
 }
 
+#[derive(Default)]
+struct SpotRequestResult<'a> {
+    id_to_name: HashMap<String, &'a str>,
+    setup_fns: HashMap<
+        &'a str,
+        &'a Box<Fn(&mut ssh::Session) -> Box<Future<Item = (), Error = failure::Error>>>,
+    >,
+    usernames: HashMap<&'a str, &'a str>,
+    spot_req_ids: Vec<String>,
+}
+
 /// Use this to prepare and execute a new tsunami.
 ///
 /// A tsunami consists of one or more [`MachineSetup`](struct.MachineSetup.html)s that will be
@@ -398,36 +409,94 @@ impl TsunamiBuilder {
         let log = &this.log;
         let expected_num: u32 = this.descriptors.values().map(|&(_, n)| n).sum();
 
-        let all_the_things = futures::future::lazy(move || {
-            info!(log, "spinning up tsunami");
+        let all_the_things = futures::future::lazy(move || Self::configure_sg(log, ec2))
+            .and_then(move |group_id| {
+                // construct keypair for ssh access
+                Self::configure_ssh(log, ec2).map(|(pk, key_name)| (pk, key_name, group_id))
+            })
+            .and_then(move |(pk, key_name, group_id)| {
+                if this.cluster {
+                    Either::A(Self::configure_placement_groups(log, ec2))
+                } else {
+                    Either::B(futures::future::ok(None))
+                }.map(move |placement| (placement, pk, group_id, key_name))
+            })
+            .and_then(move |(placement, pk, group_id, key_name)| {
+                Self::issue_spot_requests(
+                    log,
+                    ec2,
+                    &this.descriptors,
+                    this.max_duration,
+                    placement,
+                    key_name,
+                    group_id,
+                ).map(
+                    |SpotRequestResult {
+                         id_to_name,
+                         setup_fns,
+                         usernames,
+                         spot_req_ids,
+                     }| (id_to_name, setup_fns, usernames, spot_req_ids, pk),
+                )
+            })
+            .and_then(
+                move |(id_to_name, setup_fns, usernames, spot_req_ids, pk)| {
+                    Self::wait_for_instances(log, ec2, spot_req_ids, id_to_name, &this.max_wait)
+                        .map(|(machines, instances)| {
+                            (machines, instances, setup_fns, usernames, pk)
+                        })
+                },
+            )
+            .and_then(move |(machines, instances, setup_fns, usernames, pk)| {
+                Self::setup_and_cleanup(
+                    log,
+                    ec2,
+                    machines,
+                    instances,
+                    setup_fns,
+                    usernames,
+                    pk,
+                    expected_num,
+                    handle,
+                    f,
+                )
+            });
+        Box::new(all_the_things)
+    }
 
-            // set up network firewall for machines
-            let mut group_name = String::from("tsunami_security_");
-            group_name.extend(rand::thread_rng().gen_ascii_chars().take(10));
-            trace!(log, "creating security group"; "name" => &group_name);
-            let mut req = rusoto_ec2::CreateSecurityGroupRequest::default();
-            req.group_name = group_name;
-            req.description = "temporary access group for tsunami VMs".to_string();
-            ec2.create_security_group(&req)
-                .then(|r| r.context("failed to create security group for new machines"))
-        }).and_then(move |res| {
-            let group_id = res.group_id
-                .expect("aws created security group with no group id");
-            trace!(log, "created security group"; "id" => &group_id);
+    fn configure_sg<'a>(
+        log: &'a slog::Logger,
+        ec2: &'a Box<rusoto_ec2::Ec2>,
+    ) -> Box<Future<Item = String, Error = Error> + 'a> {
+        info!(log, "spinning up tsunami");
 
-            let mut req = rusoto_ec2::AuthorizeSecurityGroupIngressRequest::default();
-            req.group_id = Some(group_id.clone());
+        // set up network firewall for machines
+        let mut group_name = String::from("tsunami_security_");
+        group_name.extend(rand::thread_rng().gen_ascii_chars().take(10));
+        trace!(log, "creating security group"; "name" => &group_name);
+        let mut req = rusoto_ec2::CreateSecurityGroupRequest::default();
+        req.group_name = group_name;
+        req.description = "temporary access group for tsunami VMs".to_string();
+        let future = ec2.create_security_group(&req)
+            .then(|r| r.context("failed to create security group for new machines"))
+            .and_then(move |res| {
+                let group_id = res.group_id
+                    .expect("aws created security group with no group id");
+                trace!(log, "created security group"; "id" => &group_id);
 
-            // ssh access
-            req.ip_protocol = Some("tcp".to_string());
-            req.from_port = Some(22);
-            req.to_port = Some(22);
-            req.cidr_ip = Some("0.0.0.0/0".to_string());
-            trace!(log, "adding ssh access to security group");
-            ec2.authorize_security_group_ingress(&req)
-                .then(|r| r.context("failed to fill in security group for new machines"))
-                .map(move |_| (group_id, req))
-        })
+                let mut req = rusoto_ec2::AuthorizeSecurityGroupIngressRequest::default();
+                req.group_id = Some(group_id.clone());
+
+                // ssh access
+                req.ip_protocol = Some("tcp".to_string());
+                req.from_port = Some(22);
+                req.to_port = Some(22);
+                req.cidr_ip = Some("0.0.0.0/0".to_string());
+                trace!(log, "adding ssh access to security group");
+                ec2.authorize_security_group_ingress(&req)
+                    .then(|r| r.context("failed to fill in security group for new machines"))
+                    .map(move |_| (group_id, req))
+            })
             .and_then(move |(group_id, mut req)| {
                 // cross-VM talk
                 req.from_port = Some(0);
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -398,36 +409,94 @@ impl TsunamiBuilder {
...
+                    placement,
+                    key_name,
+                    group_id,
+                ).map(
+                    |SpotRequestResult {
+                         id_to_name,
+                         setup_fns,
+                         usernames,
+                         spot_req_ids,
+                     }| (id_to_name, setup_fns, usernames, spot_req_ids, pk),
```
jonhoo: I'd probably just pass along the entire `SpotRequestResult` here, and then deconstruct in the next future.

src/lib.rs
```diff
@@ -438,338 +507,521 @@ impl TsunamiBuilder {
...
-                let mut spot_req_ids = Vec::new();
+            .collect();
+
+        let future = futures::future::join_all(all)
+            .map(move |res_v| {
+                let mut result = SpotRequestResult::default();
                 for (name, setup, user, res) in res_v {
-                    setup_fns.insert(name, setup);
-                    usernames.insert(name, user);
+                    result.setup_fns.insert(name, &setup);
```
jonhoo: I don't think the `&` is necessary here?

src/lib.rs
```diff
@@ -438,338 +507,521 @@ impl TsunamiBuilder {
...
+        Box::new(future)
+    }
 
-                                let any_pending = res.spot_instance_requests
+    fn wait_for_instances<'a>(
+        log: &'a slog::Logger,
+        ec2: &'a Box<rusoto_ec2::Ec2>,
+        spot_req_ids: Vec<String>,
+        id_to_name: HashMap<String, &'a str>,
+        max_wait: &'a Option<time::Duration>,
```
jonhoo: I think you should just take an `Option<time::Duration>` here. Copying it seems fine.

src/lib.rs
```diff
@@ -438,338 +507,521 @@ impl TsunamiBuilder {
...
+    ) -> Box<Future<Item = (HashMap<&'a str, Vec<Machine>>, Vec<String>), Error = Error> + 'a> {
+        // Helper enum for intermediate state
+        enum WaitForInstancesResult<'a> {
+            /// instances, id_to_name
+            Ok(Vec<String>, HashMap<String, &'a str>),
```
jonhoo: Just so the flow is easier to follow, could you make the doc comments:
```rust
enum RunningInstanceSet<'a> {
    /// All instances started correctly: holds instance ids and id_to_name
    All(Vec<String>, HashMap<String, &'a str>),
    /// Some instances were not started: holds instance ids for those that were
    Only(Vec<String>),
}
```

src/lib.rs
```diff
@@ -438,338 +507,521 @@ impl TsunamiBuilder {
...
+                                match instance {
+                                    rusoto_ec2::Instance {
+                                        instance_id: Some(instance_id),
+                                        instance_type: Some(instance_type),
+                                        private_ip_address: Some(private_ip),
+                                        public_dns_name: Some(public_dns),
+                                        public_ip_address: Some(public_ip),
+                                        ..
+                                    } => {
+                                        let name = id_to_name[&instance_id].clone();
```
jonhoo: I don't think it should be necessary to `clone` here? Just stick to `&*id_to_name[&instance_id]`.

src/lib.rs
```diff
@@ -438,338 +507,521 @@ impl TsunamiBuilder {
...
+                                        return futures::future::Loop::Continue(id_to_name);
+                                    }
                                 }
-                                Some(e) => Either::B(futures::future::err(e)),
                             }
+                        }
+
+                        futures::future::Loop::Break(machines)
+                    })
+                }).map(|machine| (machine, term_instances))
```
jonhoo: Hmm, I don't think this will work? `WaitForInstancesResult::NotOk` would be turned into an `Err` around line 828, which would mean that any instances contained within it would not be mapped up here, which means that they wouldn't be terminated!

src/lib.rs
```diff
@@ -438,338 +507,521 @@ impl TsunamiBuilder {
...
-                                    // TODO: include both errors?
-                                    e
-                                },
-                            ))
-                        } else {
-                            Either::B(futures::future::err(e))
-                        }.map(|_| ())
-                            .map_err(Error::from)
+            .and_then(move |machines| {
+                let machines: HashMap<&str, _> = machines.into_iter().fold(
```
jonhoo: You likely won't need the explicit type annotation here if you make the change to `name` I suggested above.

src/lib.rs
```diff
@@ -438,338 +507,521 @@ impl TsunamiBuilder {
...
+                    HashMap::new(),
+                    |mut machines, (name, machine)| {
+                        machines.entry(&name).or_insert_with(Vec::new).push(machine);
```
jonhoo: The `&` here would also go away.

src/lib.rs
```diff
@@ -438,338 +507,521 @@ impl TsunamiBuilder {
...
+                    })
+                    .map(move |r| (r, start))
+            })
+            .map(move |(r, start)| {
+                info!(log, "the power of the tsunami was unleashed"; "duration" => start.elapsed().as_secs());
+                r
+            })
+            .or_else(move |e| {
+                // 5. terminate all instances
+                if !&term_instances.is_empty() {
```
jonhoo: You shouldn't need the `&` here; `.` automatically does dereferencing if necessary.

src/lib.rs
```diff
@@ -161,6 +161,17 @@ impl MachineSetup {
     }
 }
 
+#[derive(Default)]
+struct SpotRequestResult<'a> {
```
jonhoo: Can you also derive `Debug` and `Clone` here? It's not currently necessary, but those are nice to have :)

jonhoo:
This is great! I left some relatively minor comments inline, but those should be relatively straightforward to address :) There is one slightly bigger issue, and that is (as I comment inline) that if only *some* instances start correctly, whereas others do not, those instances will not be terminated (because the `or_else` with the termination code won't be run). In the PR as it stands, termination *only* happens if all the instances start correctly.

src/lib.rs
ViktorHavrysh: I can't derive debug because the `Fn` doesn't implement it for some reason. I can implement `Clone`, though

src/lib.rs
ViktorHavrysh: I'm not sure how to resolve this. I'm actually not even sure how it works in the original codebase. I thought I just inlined a closure, but it seems like that changed the behavior somehow?

ViktorHavrysh:
I addressed most of your concerns, but I'm not sure what to do with the `term_instances` one. I'm not sure I understand what the original code is doing, to be honest.

src/lib.rs
jonhoo: Ohh, right, that's too bad. I suppose we could implement `Debug` manually on it, but that should be dealt with in a separate PR if at all :)

ViktorHavrysh:
Actually, doesn't the `at_the_end` closure only run in the final `or_else` in the current version of the code as well?

jonhoo:
Let me see if I can explain... First, if the timer expires on the setup requests, we produce an `Ok` response that contains the instances that *were* started [here](https://github.com/jonhoo/tsunami/blob/async/src/lib.rs#L701). That means that we *do* enter the `and_then` [here](https://github.com/jonhoo/tsunami/blob/async/src/lib.rs#L742). `at_the_end` is the constructed [here](https://github.com/jonhoo/tsunami/blob/async/src/lib.rs#L746), which means that it will hold the set of instances regardless of whether we timed out or not. Only *after* that do we make the deadline expiring [into an error](https://github.com/jonhoo/tsunami/blob/async/src/lib.rs#L792). This ensures that when `at_the_end` is called, it will know about any running instances regardless of whether the spot requests timed out or not.

And yes, you seem to be totally right! The current `async` code will actually *not* terminate the instances if no error occurred :smile:, because `at_the_end` is *only* invoked in the `or_else` condition [here](https://github.com/jonhoo/tsunami/blob/async/src/lib.rs#L965). It should instead be invoked under a `then` condition so that it always triggers!

src/lib.rs
```diff
@@ -821,14 +820,13 @@ impl TsunamiBuilder {
...
-            if let WaitForInstancesResult::Ok(instances, id_to_name) = wait_result {
+        }.and_then(|wait_result| match wait_result {
+            RunningInstanceSet::All(instances, id_to_name) => {
                 Either::A(futures::future::ok((id_to_name, instances)))
-            } else {
-                Either::B(futures::future::err(format_err!(
-                    "not all instances were launched within the time limit"
-                )))
             }
+            RunningInstanceSet::Only(_) => Either::B(futures::future::err(format_err!(
```
jonhoo: Yeah, notice how the instance list is dropped here. It needs to somehow be communicated to the cleanup code so that any instances there will be terminated!

ViktorHavrysh:
After a long battle with the compiler, I think I fixed the problem with `term_instances`. But I had to refactor error handling in order to do so, and I'm not sure that it turned out better than the initial monolith :)

ViktorHavrysh:
Oh, also, in the process of refactoring I turned this:

```rust
match machine
    .public_ip
    .parse::<IpAddr>()
    .context("machine ip is not an ip address")
```

into this:

```rust
let addr = machine
    .public_ip
    .parse::<IpAddr>()
    .expect("machine ip is not an ip address");
```

I the end, this change wasn't really needed, but do we really want to handle the case when an IP is not an IP?

jonhoo:
I think this still looks a lot nicer than the old code -- great work! I have some minor comments I'll make inline, but I think we're almost there :D The extra error type is a little unfortunate, but I see why you did it. It seems like a reasonable approach, and no-one has ever complained about *more* expressive errors :p

As for `.expect` on `public_ip`, I agree that it's probably fine to assume that the value will be valid if it is given. I think my concern is more if the value is *empty*, but maybe we're guaranteed that that won't happen? In either case I think the change is okay.

src/lib.rs
```diff
@@ -164,14 +164,72 @@ impl MachineSetup {
...
+
+    #[fail(display = "failed to describe spot instances after launch: {:?}", error)]
+    DescribeWithInstances {
+        error: rusoto_ec2::DescribeInstancesError,
+        instances: Vec<String>,
+    },
+
+    #[fail(display = "tsunami main routine failed")]
+    MainRoutine {
+        error: Box<Error>,
```
jonhoo: I don't think you should need to box this? `failure::Error` is a concrete type, not a trait, and can be stored directly.

src/lib.rs
```diff
@@ -164,14 +164,72 @@ impl MachineSetup {
...
+        HashMap<&'a str, &'a Box<Fn(&mut ssh::Session) -> Box<Future<Item = (), Error = Error>>>>,
     usernames: HashMap<&'a str, &'a str>,
     spot_req_ids: Vec<String>,
 }
 
+#[derive(Debug, Fail)]
+enum TsunamiError {
+    #[fail(display = "failed to create security group for new machines: {:?}", error)]
+    SecurityGroupCreate {
+        error: rusoto_ec2::CreateSecurityGroupError,
```
jonhoo: I would make all the things that just have a single error in them be single-element tuples instead so you don't have to write `error:` everywhere :) You [can use `_0`](https://boats.gitlab.io/failure/derive-fail.html#tuple-structs) to refer to the error in `#[fail(display ... )]`.

jonhoo: Another reason to do this is that you can then just do:
```rust
.map_err(TsunamiError::SecurityGroupCreate)
```
Without the extra closure there.

src/lib.rs
```diff
@@ -669,9 +741,9 @@ impl TsunamiBuilder {
                             && msg.contains("does not exist")
                         {
                             trace!(log, "spot instance requests not yet ready");
-                            return Ok(futures::future::Loop::Continue(id_to_name));
+                            return Ok(future::Loop::Continue(id_to_name));
                         } else {
-                            return Err(e).context(format!("failed to describe spot instances"));
+                            return Err(TsunamiError::Describe);
```
jonhoo: I think you're missing the error in `e` here?

src/lib.rs
```diff
@@ -993,49 +1058,56 @@ impl TsunamiBuilder {
...
-                    ec2.delete_security_group(&req)
-                        .context("failed to clean up security group")?;
-                    trace!(log, "cleaning up temporary keypair");
-                    let mut req = rusoto_ec2::DeleteKeyPairRequest::default();
-                    req.key_name = key_name;
-                    ec2.delete_key_pair(&req)
-                    .context("failed to clean up key pair")?;
-                    // TODO: clean up created placement group
-                    */
+                ()
```
jonhoo: You shouldn't need the explicit `()` here. Statements that end with `;` evaluate to `()`, so the `info();` above should be sufficient.

jonhoo:
It's showing my two last comments as "outdated", but I don't think that's the case..? :/

ViktorHavrysh:
Yeah, that was strange. But they should really be outdated now :)

jonhoo:
Fantastic! Thank you so much for sticking it out through all my comments. I think the change is looking really good now, so happy to merge :tada: Hope you feel like the experience was helpful too!

ViktorHavrysh:
Yeah, I've learned some things, so it's been great :) Hope you'll be continuing your live-coding series.

jonhoo:
Also pointed others at this so they can see how we cleaned up the code: https://twitter.com/Jonhoo/status/983024164817833985

EDIT: And yes, the plan is definitely to continue! I have a big deadline coming up in a few weeks, but once I'm past that, I'll be back on schedule :) If you have ideas for things you'd like to see covered, feel free to ping me here, on [Twitter](https://twitter.com/Jonhoo), or on [Patreon](https://www.patreon.com/jonhoo)!

jonhoo:
I can also confirm that this now all works with my AWS setup following some [minor changes](https://github.com/jonhoo/tsunami/compare/4860911%5E...async)!

ViktorHavrysh:
Rust is the only language I'm familiar with that I'm not very surprised about that fact after that big of a refactor :D

PR #21, opened 2018-08-19
Allow setting EC2 availability zone

ms705:
This adds `set_availability_zone` on `TsunamiBuilder`, which callers can use to set the EC2 availability zone (which is more fine-grained than an EC2 "region": e.g., AZ `us-east-1a` vs. `UsEast1`) for the spot instances to be launched in.

By default, the AZ is `None` and EC2 chooses it automatically.

This affects the `availability_zone` field in `rusoto`'s [`SpotPlacement`](https://rusoto.github.io/rusoto/rusoto_ec2/struct.SpotPlacement.html) structure. Even though the documentation claims that this fields affect spot fleets only, it appears to affect even instances launched outside a spot fleet.

Note that the availability zone is passed as a string, and no validation against the set `Region` takes place, so it is currently possible to set a mismatching region and AZ (e.g., `Region::UsEast1` vs. AZ
us-west-1a").

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -153,6 +153,7 @@ pub struct TsunamiBuilder {
     log: slog::Logger,
     max_duration: i64,
     region: Region,
+    availability_zone: Option<String>,
     cluster: bool,
     max_wait: Option<time::Duration>,
 }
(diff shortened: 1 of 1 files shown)


jonhoo:
In `0.7.0`!

PR #22, opened 2019-09-26
Support multiple providers and regions

akshayknarayan:
Note: this PR changes the tsunami API significantly.

This PR adds support for VM deployments across cloud providers and cloud provider regions. That is, the following is now possible:

```
...
az.add("azure_uswest", ...);
aws.add("ec2_useast", ...);
let az = az.spawn()?;
let aws = aws.spawn()?;
let vms = az.get_machines().extend(aws.get_machines());
...
```

To support this, the existing functionality is moved to src/providers/aws.rs, and put behind a `Launcher` trait, which is a common interface across cloud providers.

An Azure provider is implemented (src/providers/azure.rs), as is one that just wraps a connection to an existing bare-metal machine (src/providers/baremetal.rs).

PR #23, opened 2020-04-02
Support multiple providers and regions

akshayknarayan:
This is a cleaned-up version of https://github.com/jonhoo/tsunami/pull/22.


znewman01:
FYI I might be able to test out this PR this weekend, will report back.

znewman01:
Been using this, though not yet successfully. Here's my thoughts (do with them
what you will, up to and including ignoring them entirely).

(Happy to talk in more detail on Zoom etc. if you want)

## Blockers

- `Launcher::default()` should definitely not create a runtime. I'm using this
  from within an async context and getting panics.
  Honestly, it's much better IMO to just make the API async and not worry about
  runtime at all.
  
## Ergonomics / Design

- The object model seems a little confused, and you hew a little closely to
  the first iteration of Tsunami.
  
  Specifically:
  
  - running `<thing where I specified all the machines>.spawn(<thing that does the launching>)`
    seems backwards
  - why do I have to manage the launcher myself?
  - `Setup` is not a meaningful name
  - why do I go back to the launcher to get the SSH connections?
  - why is spawning instances separate from making SSH connections to them?
  - I declared 10 VMs nicknamed "worker", why do I now have to look for the name
    "worker-..." in the result of `connect_all()`?

  Spitballing:
  
  - I'd want everything to be declarative and immutable up until the point that
    you run `spawn()` (ex. be able to reuse a `Launcher`).
  - `Launcher`: pretty much what it is now, less the state for managing machines
  - `Launcher::LaunchSpec`: a specification of what machines (can handle
    grouping by region, etc.) to launch with this `Launcher`
    - `Launcher::launch(Self::LaunchSpec)`
    - ex. `aws::Launcher::LaunchSpec = HashMap<aws::Machine>` (or a thin wrapper)
  - `Tsunami`: a wrapper around a set of `(Launcher, Launcher::LaunchSet)`
    (you'd need a bunch of `Box`es to make this work probably).
    - `spawn` would spawn and connect to all VMs for you
    - and return
      ```rust
      struct SpawnResult {
         machine: HashMap<String, Machine>,
         groups: HashMap<String, Vec<Machine>>,
         /* private stuff */
      }
      ```
      where the `Drop` terminator was related to the lifetime of the
      `SpawnResult`

    
  Would look something like:

  ```rust
  let tsunami = TsunamiBuilder()::default()
      .add_launch_set(
          aws::Launcher::new(/* credential info */),
          aws::LaunchSpec::default()
              .add(
                  "worker",
                  aws::Machine::default()
                     .instance_type("most expensive")
              )
      )
      .add_launch_set(
          azure::Launcher::default(),
          azure::LaunchSet::default()
              .add_multiple(
                  10,
                  "client",
                  azure::Machine::default()
                     .region(France)
              )
      )
  ):
  
  for trial in 0..5 {
    let vms = tsunami.spawn()?

    vms.machines["worker"].ssh(/* ... */);
    for vm in vms.groups["clients"] {
        vm.ssh(/* ... */);
    }
  }
  ```
  
  This has some nice consequences:
  
  - everything is reusable
  - for `baremetal`, we can make the `LaunchSet` type necessarily only one machine
  - I don't really think about launchers after creating them once
  - I don't need to manage multiple `Tsunami`s
  - 
  

- `region_with_ubuntu_ami` is a little stilted.
  Similarly, the `Setup<Yes/No>` trick is cute but the error messages aren't useful.
  Really what you're trying to ensure here is: "if someone passes a custom AMI
  in, they better give us a username too." Proposal:
  ```rust
  enum Ami {
    /// Use the default Ubuntu 18.04 AMI for the provided region
    struct Ubuntu,
    /// Provide an AMI, along with the username of the default user.
    struct Custom {
       id: String,
       username: String
    }
  }
  
  impl Setup {
     pub fn ami(ami: Ami) -> Setup;
  }
  ```
  This makes invalid states unrepresentable via the type system.
  
## Cloud Stuff

- Hard-coding AMIs seems sketchy to me---they're already quite out of date.
  Might make sense to query (somebody?) for these.
  
- If you're worried about shutting down Azure instances if we crash (a valid
  worry!), maybe make an Azure function and hook it up to a 1-time timer
  
- I'm personally insulted by the lack of GCP support :)

## Misc. ideas/thoughts

- Gate the various providers behind Crate features? I feel silly watching Azure stuff
  compile if I'm not using it.

- Using Rayon in addition to Tokio now feels silly, especially since we're not
  doing anything compute heavy.

- I think "real work in destructor" is a code smell (doubly so when that work is async!). I really like the closure method from the original API, and I think that would sidestep the issue nicely.

znewman01:
(also I'm happy to work on some of these changes if you want, I'm running some experiments this week)

znewman01:
FYI [znewman01/tsunami 2a53f32](https://github.com/znewman01/tsunami/commit/2a53fe2076f343ee475f2048b988e5b0d3dd57bf) does the async stuff I need, a little hacky

jonhoo:
Wow, thanks for that — I'll try to make my way through your comment once I get a chance. I'm sure @akshayknarayan also has thoughts! One quick note: I _really_ want tsunami's API to be async. The main reason it is not at the moment is that there is no library out there that really does asynchronous SSH (at least not last time I looked), so ultimately we will end up doing blocking calls. But maybe we should just bite the bullet and use `block_in_place` for those and make everything else async?

znewman01:
>  But maybe we should just bite the bullet and use block_in_place for those and make everything else async?

Yeah: I think that's totally worth doing now, as it'll be the least disruptive later when you (hopefully/eventually) swap out the underlying library, as long as you warn in the docs. I think the benefits outweigh the drawbacks (internally and for end users).

(But maybe not this PR :smile:  )

src/providers/aws.rs
```diff
@@ -0,0 +1,1075 @@
...
+            .context("failed to create security group for new machines")?;
+        let group_id = res
+            .group_id
+            .expect("aws created security group with no group id");
+        trace!(log, "created security group"; "id" => &group_id);
+
+        let mut req = rusoto_ec2::AuthorizeSecurityGroupIngressRequest::default();
+        req.group_id = Some(group_id.clone());
+
+        // icmp access
```
znewman01, line 701: how about ssh? :)

akshayknarayan: ssh is covered by the blanket tcp rule below this. Having a blanket tcp rule is suboptimal, but that is an issue for another PR I think.

akshayknarayan:
I definitely agree with @jonhoo that it would be great if the API was async - your comment about running Launcher::default() from within an async runtime is reasonable.
The rayon point is valid, but this is also related to the ssh library - since it is blocking, there can only be as many active ssh connections as the number of cores, so `async` isn't really helping.  

re: API
I am not particularly attached to the current API, and your observation that it hews closely to the existing one is apt - I did try to minimize changes.
The idea of the current structure is that you spawn machines onto a launcher, then access those machines accordingly. @jonhoo and I actually initially tried a version of this where machines are spawned the opposite way, and tied to the lifetime of the `Tsunami`.
iirc, we went down the machines-go-into-the-launcher route to avoid having to have a `Vec<Box<dyn Provider>>` (as you mention) or similar in the `TsunamiBuilder`. I'm not saying this is bad, but it is something to consider.

re: your other points:
- I have been meaning to do GCP support, but haven't had time :(
- Your idea of using Azure functions to shut down instances via timer is a good one - do you know if it is possible to do this entirely via the azure CLI, since that is the interface we have?
- Gating providers behind features is a good idea
- It is probably possible to scrape AMIs from https://ec2instances.info/, but that seems like an issue for another PR. The current goal is just to provide some default.

znewman01:
First, hopefully that didn't come off too critical! I'm happily using the PR so this was just meant as constructive feedback.

> The rayon point is valid, but this is also related to the ssh library - since it is blocking, there can only be as many active ssh connections as the number of cores, so `async` isn't really helping.

Yeah, but you can imagine doing e.g. the Rusoto requests async while the SSH bits remain sync. And if we're still using the Azure CLI, `tokio::process` is a pretty-much drop-in replacement for `std::process`. I think that part would be easier to express async. 

> iirc, we went down the machines-go-into-the-launcher route to avoid having to have a `Vec<Box<dyn Provider>>` (as you mention) or similar in the `TsunamiBuilder`. I'm not saying this is bad, but it is something to consider.

Oh for sure, I think it might cause some headaches internally. But IMO the ergonomics are really worth it.

> I have been meaning to do GCP support, but haven't had time :(

Oh, I'm mostly teasing--I just used to work on on GCP :)

>  Your idea of using Azure functions to shut down instances via timer is a good one - do you know if it is possible to do this entirely via the azure CLI, since that is the interface we have?

My Azure experience is limited but I know it would have been possible using the GCP CLI :)
Seems like the CLI [does support functions](https://docs.microsoft.com/en-us/azure/azure-functions/functions-create-first-azure-function-azure-cli?tabs=bash%2Cbrowser&pivots=programming-language-csharp) but you have to install "Azure Functions Core Tools" first.

Bummer that [MindFlavor/AzureSDKForRust](https://github.com/MindFlavor/AzureSDKForRust) doesn't support more services. It seems actively maintained and usable.

> It is probably possible to scrape AMIs from https://ec2instances.info/, but that seems like an issue for another PR. The current goal is just to provide some default.

Yeah, I just wanted to braindump all the thoughts I had. Feel free to totally disregard 
(except you probably should fix SSH for AWS :) ).

src/providers/aws.rs
znewman01: Unfortunately, if `use_open_ports` is false (which seems to be both the default and what you actually want), this errors out during machine setup. We can address later if that's more convenient but it's not working right now

znewman01:
@jonhoo  have you seen https://nest.pijul.com/pijul_org/thrussh ? It's a little young but might fit the bill

jonhoo:
[Yup](https://www.youtube.com/watch?v=RBQwZthJjoM&list=PLqbS7AVVErFjzlF3JSwzxttE1G0awWimV) ;) My experience from back then was that it was pretty low level and still not quite usable in real-world situations, but that may well have changed since. If someone wants to take a stab at getting it to work here, I'm all for it!

src/providers/aws.rs
```diff
@@ -0,0 +1,1075 @@
...
+        let private_key_path = self.private_key_path.as_ref().unwrap();
+        let mut all_ready = self.instances.is_empty();
+        desc_req.instance_ids = Some(self.instances.keys().cloned().collect());
+        while !all_ready {
+            all_ready = true;
+
+            for reservation in client
+                .describe_instances(desc_req.clone())
+                .await
+                .context("failed to cancel spot instances")?
```
znewman01: bad err context

src/providers/aws.rs
akshayknarayan: hm yeah, `use_open_ports` is necessary for ssh. I think it is worth adding a special rule for ssh as @znewman01 mentioned - @jonhoo thoughts?

akshayknarayan:
> First, hopefully that didn't come off too critical! I'm happily using the PR so this was just meant as constructive feedback.

Not at all! It's always useful to get feedback, and a better tsunami means an easier time writing experiments for all of us :) 

> Oh, I'm mostly teasing--I just used to work on on GCP :)

btw, do you happen to know of a GCP rust api library? :)

> Bummer that MindFlavor/AzureSDKForRust doesn't support more services. It seems actively maintained and usable.

Yeah, I've used it before and it is pretty reasonable - I wish they supported more of azure, but iirc it's a side project by a microsoft employee so my hopes aren't high.

Overall, I'm generally in favor of your other comments re: async api (`Vec<Box<dyn Provider>>` I am on the fence about), but I think this PR is itself a big enough change that they should be addressed separately

src/providers/aws.rs
akshayknarayan: I added this (b61ef49)

znewman01:
Thanks for the follow-up!

I ultimately added an  [OnDemandLauncher](https://github.com/znewman01/tsunami/commit/8c91e80513f6c587621d2ac30a88e8ed046e2430) to AWS, as I wasn't finding enough spot capacity for my needs. It's a huge hack but at some point I'll clean it up and send a PR.
 
> btw, do you happen to know of a GCP rust api library? :)

Definitely nothing official. There's an [autogenerated one](https://github.com/Byron/google-apis-rs) which should be adequate; I used to work on the CLI which used the official auto-generated Python bindings (via [apitools](https://github.com/google/apitools)) and we never had issues; it's a very similar experience to using Rusoto.

> Overall, I'm generally in favor of your other comments re: async api (`Vec<Box<dyn Provider>>` I am on the fence about), but I think this PR is itself a big enough change that they should be addressed separately

Yeah, doing later makes sense. Happy to spend an hour or two sketching out an API with you and Jon when (if?) we're back in Stata if that'd be useful.

jonhoo:
@znewman01 Now that I've started writing up https://github.com/jonhoo/openssh-rs/, we may even have async ssh sooner than anticipated. Should be pretty easy to provide async types in that I think, since it's really just communicating with local processes, which can be encapsulated with https://docs.rs/tokio/0.2/tokio/process/index.html. Not entirely sure yet what the right way to provide both async and non-async modes is, but even _just_ having async for `STDIN` and `STDOUT` should go a long way.

jonhoo:
Okay, so, I started porting one of the Noria orchestration programs now, and some notes:

 - It looks like it's no longer possible to set the AWS Availability Zone?
 - The errors (sadly) get really confusing with the `Yes` and `No` marker traits, since `rustc` does not also report which of the generic types there's a problem with. The errors end up looking like:
   ```
   error[E0277]: the trait bound `tsunami::providers::aws::Setup<tsunami::providers::aws::Yes, tsunami::providers::aws::No>: tsunami::providers::MachineSetup` is not satisfied
      --> orchestration/src/lobsters-noria.rs:111:17
       |
   111 |     let mut b = TsunamiBuilder::default();
       |                 ^^^^^^^^^^^^^^^^^^^^^^^ the trait `tsunami::providers::MachineSetup` is not implemented for `tsunami::providers::aws::Setup<tsunami::providers::aws::Yes, tsunami::providers::aws::No>`
       |
       = help: the following implementations were found:
                 <tsunami::providers::aws::Setup as tsunami::providers::MachineSetup>
   ```
   Which is pretty hard to understand. I wonder if we need to introduce type aliases for `Yes` and `No` like `HasUsername` and `HasAMI`...
 - I think we'll want an `openssh` release of https://github.com/akshayknarayan/sessh. It might help if `openssh` itself just adopted `cmd` I suppose, but `sessh` provides other conveniences too. You should be able to mostly copy-paste the changes from https://github.com/jonhoo/tsunami/compare/openssh.
 - I was pretty surprised that I couldn't do:
   ```rust
   aws = aws.with_credentials(move || {
       let sts = rusoto_sts::StsClient::new(rusoto_core::Region::UsEast1);
       rusoto_sts::StsAssumeRoleSessionCredentialsProvider::new(
           sts,
   	// ...
       )
   });
   ```
   because the method expects a closure that returns a `Result`. Are we expecting the closure to error often? Can we just make it return a `P2` or whatever it is?
 - This error was pretty confusing to me:
   ```
   error[E0599]: no method named `connect_all` found for struct `tsunami::providers::aws::Launcher<rusoto_sts::custom::credential::StsAssumeRoleSessionCredentialsProvider>` in the current scope
      --> orchestration/src/lobsters-noria.rs:221:23
       |
   221 |     let mut vms = aws.connect_all().unwrap();
       |                       ^^^^^^^^^^^ method not found in `tsunami::providers::aws::Launcher<rusoto_sts::custom::credential::StsAssumeRoleSessionCredentialsProvider>`
   ```
   The compiler was helpful enough to point out that I need to `use tsunami::providers::Launcher;`, but may be good calling this out explicitly in the documentation.
 - `aws::Machine` is missing `private_ip`.

Given ^ I haven't had a chance to actually _run_ the thing yet, but it's a start!

jonhoo:
Two more things:

 - I **love** being able to get rid of the `run_as` closure!
 - While writing the code, I started wondering what would happen if I accidentally called `spawn` with the same launcher more than once...

akshayknarayan:
>     * It looks like it's no longer possible to set the AWS Availability Zone?

#29 :)

>     * The errors (sadly) get really confusing with the `Yes` and `No` marker traits, since `rustc` does not also report which of the generic types there's a problem with. The errors end up looking like:

Yeah, @znewman01 brought this up too. We should probably simplify the api to just force specifying relevant arguments together.
 
>     * I think we'll want an `openssh` release of https://github.com/akshayknarayan/sessh. It might help if `openssh` itself just adopted `cmd` I suppose, but `sessh` provides other conveniences too. You should be able to mostly copy-paste the changes from https://github.com/jonhoo/tsunami/compare/openssh.

Hm, I was actually thinking `sessh` could mostly go away - the main things `sessh` provides are `cmd` + file upload and download, which `openssh` now has.
 
>       because the method expects a closure that returns a `Result`. Are we expecting the closure to error often? Can we just make it return a `P2` or whatever it is?

The original reason was that https://docs.rs/rusoto_credential/0.43.0/rusoto_credential/struct.DefaultCredentialsProvider.html#method.new returns a `Result`, but if we are ok with making people `unwrap` then it could return a `P2`.

>       The compiler was helpful enough to point out that I need to `use tsunami::providers::Launcher;`, but may be good calling this out explicitly in the documentation.

Where in the documentation were you thinking? Maybe as a comment in the doc-examples, something like this:

```
// Note that this is necessary to get ssh access to the machines, via `connect_all()`
use tsunami::providers::Launcher;
```
 
>     * `aws::Machine` is missing `private_ip`.

There is no `aws::Machine`, just `crate::Machine`? `connect_all` returns `crate::Machine`. We could potentially add an `private_ip: Option<_>` field there.

jonhoo:
> We should probably simplify the api to just force specifying relevant arguments together.

Yeah, that seems reasonable to me.

> Hm, I was actually thinking sessh could mostly go away - the main things sessh provides are cmd + file upload and download, which openssh now has.

Yeah, that's a good point. I guess we mostly just need `cmd` to be added.

> but if we are ok with making people unwrap then it could return a P2.

Ah, I missed that that was the case for `DefaultP`. In that case, `Result` seems fine. Better to `Ok`-wrap than force an unwrap.

> Maybe as a comment in the doc-examples, something like this:

That seems reasonable!

> connect_all returns `crate::Machine`. We could potentially add an `private_ip: Option<_>` field there.

Ah, sorry, yes, in whatever `Machine` it is we end up returning. Are there cases where a private ip is _not_ provided?

akshayknarayan:
> Ah, sorry, yes, in whatever `Machine` it is we end up returning. Are there cases where a private ip is _not_ provided?

The baremetal provider obviously has no sense of a private ip :) Take a look at `24827a6`? (I haven't figured out how to populate azure's private ip field yet).

src/providers/aws.rs
```diff
@@ -957,10 +988,19 @@ impl RegionLauncher {
...
             .values()
             .map(|info| match info {
-                (Some((public_ip, public_dns)), (name, Setup { username, .. })) => {
+                TaggedSetup { 
+                    name,
+                    setup: Setup { username, .. },
+                    ipinfo: Some(IpInfo {
+                        dns: public_dns,
+                        pub_ip: public_ip,
+                        priv_ip: private_ip,
```
jonhoo: Why not just name these with the long names so you wouldn't have to do things like this? :p

src/providers/azure.rs
```diff
@@ -313,6 +313,7 @@ impl super::Launcher for RegionLauncher {
                     nickname: name.clone(),
                     public_dns: ip.clone(),
                     public_ip: ip.clone(),
+                    private_ip: None,
```
jonhoo: I'm surprised that Azure VMs do not also have internal IPs?

src/providers/baremetal.rs
```diff
@@ -190,6 +190,7 @@ impl super::Launcher for Machine {
             nickname: self.name.clone(),
             public_dns: addr.to_string(),
             public_ip: addr.ip().to_string(),
+            private_ip: None,
```
jonhoo: I think this should probably be set to the same value as the `public_ip`.

jonhoo:
Left notes _before_ reading your comment, sorry about that :p
Also, I think we should just fold #28 into this, rather than keeping it separate and rebasing and stuff. I agree that normally it should be separate, but here I think we're making enough changes that folding it in isn't that egregious.

src/providers/aws.rs
akshayknarayan: fair enough :)

src/providers/azure.rs
akshayknarayan: It does - I have to figure out how to query them via the cli.

src/providers/baremetal.rs
jonhoo: nvm

README.md
```diff
@@ -1,50 +1,73 @@
-# tsunami
-
-[![Crates.io](https://img.shields.io/crates/v/tsunami.svg)](https://crates.io/crates/tsunami)
-[![Documentation](https://docs.rs/tsunami/badge.svg)](https://docs.rs/tsunami/)
```
jonhoo, line 4: Huh, why did these two go away?

README.md
```diff
@@ -1,50 +1,73 @@
...
 
 ## Live-coding
 
-The crate is under development as part of a live-coding stream series intended for users who
+An earlier version of this crate was written as part of a live-coding stream series intended for users who
 are already somewhat familiar with Rust, and who want to see something larger and more involved
 be built. You can find the recordings of past sessions [on
 YouTube](https://www.youtube.com/playlist?list=PLqbS7AVVErFgY2faCIYjJZv_RluGkTlKt).
+
+License: MIT/Apache-2.0
```
jonhoo: We probably should include the full text as suggested by the [API guidelines](https://rust-lang.github.io/api-guidelines/necessities.html#c-permissive).

README.md
akshayknarayan: ok, they should be back now

jonhoo:
I'm re-writing a pretty large orchestration thing now, and just noticed that we probably need https://github.com/jonhoo/tsunami/blob/c568ae1fb1438b0a68f0f720b880aa339cc42527/src/providers/aws.rs#L286
to be `Send + Sync`. Should just be a matter of adding that to the bound (and to the bound of `with_credentials`).

jonhoo:
Oh, this also doesn't have the `openssh` changes I think?

jonhoo:
A couple of other observations:

 - I found it pretty hard to navigate what all the different methods do, and which things I have to know about and which things I don't. For example, a bunch of the types and methods under `providers` are "internal", and most users won't need them. Similarly, only really the `spawn` method on `Launcher` is of interest to the user, and the other methods on there the user probably won't really care about. I wonder if we can shift things around so that "things you'll likely" want is better delineated. For example, `Launcher` could probably be split so that the required methods (which the user likely won't call, or even need to know about) are on a new, non-public trait, and `Launcher` _just_ has the user-facing method, and has a blanket impl for any type that implements the internal trait.
 - The doctests don't seem to be `rustfmt` formatted, and they also have some unnecessary imports in them. This makes them a little awkward to copy-paste from.
 - The fact that the user has to add dependencies on `failure` (for `failure::Error`, which appears everywhere) and `slog` (for `slog::Logger`, which is an argument to setup) is pretty sad. The fix for this is probably to move to `Box<dyn Error>` (or the `anyhow` crate), and `tracing`, but that can be done in a separate PR.
 - The `Launcher` trait is currently pretty hard to discover. It'd be good for it to be re-exported (and maybe _only_ exposed in the root of the crate, since the user will likely interact with it.
 - I think it'd be worthwhile to re-export `rusoto_core::Region`.

akshayknarayan:
> I found it pretty hard to navigate what all the different methods do, and which things I have to know about and which things I don't. For example, a bunch of the types and methods under `providers` are "internal", and most users won't need them. Similarly, only really the `spawn` method on `Launcher` is of interest to the user, and the other methods on there the user probably won't really care about. I wonder if we can shift things around so that "things you'll likely" want is better delineated. For example, `Launcher` could probably be split so that the required methods (which the user likely won't call, or even need to know about) are on a new, non-public trait, and `Launcher` _just_ has the user-facing method, and has a blanket impl for any type that implements the internal trait.
> The `Launcher` trait is currently pretty hard to discover. It'd be good for it to be re-exported (and maybe _only_ exposed in the root of the crate, since the user will likely interact with it.

I think it is helpful to allow external crates to `impl Launcher` - e.g. for aws on-demand instances etc, so that in theory we won't have to have all the impls in this crate.

I think the main confusion point is between `launch()` and `spawn()`, where `spawn()` is a more convenient version of `spawn()` that removes the need to think about `LaunchDescriptor` and does the grouping automatically. `connect_all()` and `cleanup()` are both directly user-facing. In fact, the idea of `LaunchDescriptor` is potentially annoying, since it's nice for `impl Launcher`s to not think about grouping into regions, but conversely it's an error to put machines corresponding to different regions in the same `LaunchDescriptor`.

What about this: at crate root we put `pub trait Tsunami { fn spawn(); fn connect_all(); fn cleanup(); } with a blanket `impl<L: Launcher> Tsunami for L`. That way the lower level stuff for implementors is still there and trait Tsunami is the convenient user-facing api.

jonhoo:
> What about this: at crate root we put `pub trait Tsunami { fn spawn(); fn connect_all(); fn cleanup(); } with a blanket `impl<L: Launcher> Tsunami for L`. That way the lower level stuff for implementors is still there and trait Tsunami is the convenient user-facing api.

That seems great! That top-level trait should then be sealed.

jonhoo:
It'd probably also then be good to document (on `Launcher`) when you'd want to use/implement that trait. And point other users to the `Tsunami` trait instead.

src/lib.rs
```diff
@@ -176,6 +181,96 @@ impl<'t> Machine<'t> {
...
+        log: Option<slog::Logger>,
+    ) -> Pin<Box<dyn Future<Output = Result<(), Error>> + 'l>> {
+        self.spawn(descriptors, max_wait, log)
+    }
+}
+
+mod sealed {
+    pub trait Sealed {}
+
+    impl<T: super::Tsunami> Sealed for T {}
```
jonhoo: This doesn't seem right? I think this should be `T: Launcher`?

src/providers/aws.rs
```diff
@@ -11,7 +11,8 @@
 //! ```rust,no_run
 //! #[tokio::main]
 //! async fn main() {
-//!     use tsunami::providers::{aws, Launcher};
+//!     use tsunami::Tsunami;
+//!     use tsunami::providers::{aws};
```
```suggestion
//!     use tsunami::providers::aws;
```

src/providers/aws.rs
```diff
@@ -41,7 +42,8 @@
 //! ```
 //! ```rust,no_run
 //! use rusoto_core::{credential::DefaultCredentialsProvider};
-//! use tsunami::providers::{aws::{self, Region}, Launcher};
+//! use tsunami::Tsunami;
+//! use tsunami::providers::{aws::{self, Region}};
```
```suggestion
//! use tsunami::providers::aws::{self, Region};
```

src/providers/azure.rs
```diff
@@ -18,7 +18,8 @@
 //!
 //! # Example
 //! ```rust,no_run
-//! use tsunami::providers::{azure, Launcher};
+//! use tsunami::Tsunami;
+//! use tsunami::providers::{azure};
```
jonhoo:
```suggestion
//! use tsunami::providers::azure;
```

src/providers/azure.rs
```diff
@@ -46,7 +47,8 @@
 //! }
 //! ```
 //! ```rust,no_run
-//! use tsunami::providers::{azure, Launcher};
+//! use tsunami::Tsunami;
+//! use tsunami::providers::{azure};
```
jonhoo:
```suggestion
//! use tsunami::providers::azure;
```

src/lib.rs
```diff
@@ -176,6 +181,96 @@ impl<'t> Machine<'t> {
...
+    ///     .await?;
+    ///     // access the host via the launcher
+    ///     let vms = aws.connect_all().await?;
+    ///     // we're done! terminate the instance.
+    ///     aws.cleanup().await?;
+    ///     Ok(())
+    /// }
+    /// ```
+    fn spawn<'l>(
+        &'l mut self,
```
jonhoo, line 229: I just realized that with the current signature, it's not possible to spawn some machines, then connect to them, and then spawn some more machines. Even though machines stay up for as long as the launcher lives once spawned. Not sure how to really fix this though...

src/lib.rs
```diff
@@ -176,6 +181,96 @@ impl<'t> Machine<'t> {
...
+        log: Option<slog::Logger>,
+    ) -> Pin<Box<dyn Future<Output = Result<(), Error>> + 'l>>;
+
+    /// Return connections to the [`Machine`s](crate::Machine) that `spawn` spawned.
+    fn connect_all<'l>(
+        &'l self,
+    ) -> Pin<Box<dyn Future<Output = Result<HashMap<String, crate::Machine<'l>>, Error>> + 'l>>;
+
+    /// Shut down all instances.
+    fn cleanup(self) -> Pin<Box<dyn Future<Output = Result<(), Error>>>>;
```
jonhoo: Should this be called `shutdown`? Or `terminate_all`?

src/providers/mod.rs
```diff
@@ -0,0 +1,204 @@
...
+    /// Implementors should remember enough information to subsequently answer
+    /// calls to `connect_all`, i.e., the IPs of the machines.
+    ///
+    /// This method can be called multiple times. Subsequent calls to
+    /// `connect_all` should return the new machines as well as any previously
+    /// spawned machines.
+    fn launch<'l>(
+        &'l mut self,
+        desc: LaunchDescriptor<Self::MachineDescriptor>,
+    ) -> Pin<Box<dyn Future<Output = Result<(), Error>> + 'l>>;
```
jonhoo: This should also be `+ Send + Sync`.

src/providers/mod.rs
```diff
@@ -0,0 +1,204 @@
...
+
+    /// Return connections to the [`Machine`s](crate::Machine) that `launch` spawned.
+    fn connect_all<'l>(
+        &'l self,
+    ) -> Pin<Box<dyn Future<Output = Result<HashMap<String, crate::Machine<'l>>, Error>> + 'l>>;
```
jonhoo: This should also be `+ Send + Sync`.

src/providers/mod.rs
```diff
@@ -0,0 +1,204 @@
...
+    /// Helper method to group `MachineDescriptor`s into regions and call `launch`.
+    ///
+    /// This implementation initializes each region serially. It may be useful for performance to
+    /// provide an implementation that initializes the regions concurrently.
+    fn spawn<'l>(
+        &'l mut self,
+        descriptors: impl IntoIterator<Item = (String, Self::MachineDescriptor)> + 'static,
+        max_wait: Option<std::time::Duration>,
+        log: Option<slog::Logger>,
+    ) -> Pin<Box<dyn Future<Output = Result<(), Error>> + 'l>> {
```
jonhoo: This should also be `+ Send + Sync`.

src/lib.rs
```diff
@@ -1,796 +1,330 @@
...
+    ///     aws.cleanup().await?;
+    ///     Ok(())
+    /// }
+    /// ```
+    fn spawn<'l>(
+        &'l mut self,
+        descriptors: impl IntoIterator<Item = (String, Self::MachineDescriptor)> + 'static,
+        max_wait: Option<std::time::Duration>,
+        log: Option<slog::Logger>,
+    ) -> Pin<Box<dyn Future<Output = Result<(), Error>> + 'l>>;
```
jonhoo: This should also be `+ Send + Sync`.

src/lib.rs
```diff
@@ -1,796 +1,330 @@
...
+
+    /// Return connections to the [`Machine`s](crate::Machine) that `spawn` spawned.
+    fn connect_all<'l>(
+        &'l self,
+    ) -> Pin<Box<dyn Future<Output = Result<HashMap<String, crate::Machine<'l>>, Error>> + 'l>>;
```
jonhoo: This should also be `+ Send + Sync`.

src/lib.rs
jonhoo: This should also be `+ Send + Sync`.

src/lib.rs
jonhoo: We'd have to make `spawn` take `&self`, and then use an internal `Mutex` or something to ensure that the spawns happen sequentially. Seems not worth it. And this should be pretty uncommon anyway. Feels more "right" to have it take `&mut self`. So I'll resolve for now.

jonhoo (approved):

jonhoo:
Finally, it landed! :tada:
Thanks for all your awesome work on this :heart: 
Now for some tidying up of errors and logging!

PR #24, opened 2020-04-03
Add support for copying files

znewman01:
Hi Jon! Not sure if you know me, I sit a floor down from you (err...not lately,
but you you know what I mean). Thanks for the library, it's been useful.

I've added, in the style of `scp(1)`, support for uploading/downloading files:

```rust
MachineSetup::new("c5.xlarge", "ami-e18aa89b", |ssh| {
    ssh.upload(
        Path::new("/path/on/local/machine.rar"),
        Path::new("dest/on/remote/machine.rar")
    )?;
    ssh.download(
        Path::new("remote-machine/business.xls"),
        Path::new("/tmp/business.xls")
    )?;
    Ok(())
});
```

I've verified that it works in a quick example, and I'm planning to use it for
some experiments very soon (copying a source tarball up, then building it, then
copying the binaries back down). I'll let you know if that doesn't go well.

I'm almost tempted to implement this as `Session.open(&Path) -> Result<ssh2::File, _>` to enable:

``` rust
let remote_src = ssh.open(Path::new("file"))?;
let remote_dst = ssh.open(Path::new("file"))?;
std::io:copy(remote_src, remote_dst)?;
```

but this oughta suffice for most use cases.

### DIFF
--- a/src/ssh.rs
+++ b/src/ssh.rs
@@ -1,11 +1,20 @@
 use failure::Error;
+use failure::Fail;
 use failure::ResultExt;
 use slog;
 use ssh2;
+use std::fs::File;
+use std::io;
 use std::net::{SocketAddr, TcpStream};
 use std::path::Path;
 use std::thread;
 use std::time::{Duration, Instant};
+
+#[derive(Debug, Fail)]
+enum SshError {
+    #[fail(display = "error transferring file {}: {}", file, msg)]
+    FileTransferFailure { file: String, msg: String },
+}
 
 /// An established SSH session.
 ///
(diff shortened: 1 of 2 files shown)


src/ssh.rs
```diff
@@ -1,12 +1,21 @@
...
 use ssh2;
+use std::fs::File;
+use std::io;
 use std::net::{SocketAddr, TcpStream};
 use std::path::Path;
 use std::thread;
 use std::time::{Duration, Instant};
 
+#[derive(Debug, Fail)]
+enum SshError {
```
jonhoo: This can probably be a `struct FileTransferError` rather than an `enum`, since there aren't any other variants :)

jonhoo (approved):
This looks great modulo the error change I pointed out above!

src/ssh.rs
znewman01: Done, fair :)

znewman01:
@jonhoo ready for you!

jonhoo:
Beautiful, thank you! I'm in a weird position where I want to wait for #23 before doing another (breaking) release, but I don't want to release it yet because I haven't had a chance to test it properly yet. Are you running on top of that code, and if so, can you also confirm that it works?

znewman01:
I haven't been using #23 (don't need it for my use case), though I suppose it'd be pretty easy for me to test it out (though I wouldn't be exercising the Azure code or even multiple regions within AWS). Why don't I take a crack this weekend and see if it's not too much work?

jonhoo:
That'd be awesome! Even just getting more confidence that the updated AWS code works, and that the API is okay to use would be super helpful (cc @akshayknarayan).

akshayknarayan:
btw, #23 moves the ssh functionality to a separate crate - https://github.com/akshayknarayan/sessh. It has similar functionality to what you want (https://docs.rs/sessh/0.1.3/sessh/struct.Session.html#method.scp_recv), but your convenience methods would be useful there as well!

znewman01:
oh cool, maybe I'll send a (mostly copy+paste) PR over there

PR #27, opened 2020-04-14
Allow openssh-rs backend

akshayknarayan:
Based on https://github.com/jonhoo/tsunami/pull/23, so that should go first.


jonhoo:
Hmm, this approach probably won't work, since the features won't be additive. If cargo ends up enabling both (say, because one dependency has one and another one the other), then the crate won't compile. I'm also not sure we _want_ to provide a choice. I think we should probably _just_ provide `openssh`.

akshayknarayan:
>  I'm also not sure we want to provide a choice. I think we should probably just provide openssh.

Fair enough - I'll close this and re-open one with just the `openssh` part.

PR #28, opened 2020-04-15
Switch to using openssh-rs

akshayknarayan:
Based on #23.
The diff for this PR starts at commit 279f59c.


akshayknarayan:
Folding into #23

PR #29, opened 2020-04-16
Availability zones for AWS

akshayknarayan:
Add (back :)) support for specifying aws availability zones (c18f258)


jonhoo:
I think you can just fold this in under #23 :)

PR #30, opened 2020-04-21
Spawn in launcher

akshayknarayan:
#23 has the concept of a per-provider `TsunamiBuilder`. It turns out this is not very useful and in fact sometimes annoying to have to create multiple `TsunamiBuilder`s. Get rid of it, and replace with a default implementation of `spawn` on the `Launcher` trait.

The methods on `TsunamiBuilder` become crate-level helper methods.


jonhoo:
cc @znewman01

znewman01:
IMO great change, thanks :)

will take a look probably tomorrow

src/providers/aws.rs
```diff
@@ -83,6 +83,26 @@ pub struct No;
...
 pub struct Yes;
 
+/// Available configurations of availability zone specifiers.
+///
+/// See [the aws docs](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#using-regions-availability-zones-launching) for more information.
+#[derive(Debug, Clone)]
+pub enum AvailabilityZoneSpec {
+    /// `Any` (the default) will place the instance anywhere there is capacity.
+    Any,
+    /// `Cluster` will place all the instances with the given id together, without specifying
```
znewman01: this might benefit from a little bit of elaboration--I don't totally get the "given ID"/usize thing

src/providers/aws.rs
```diff
@@ -107,14 +128,19 @@ impl super::MachineSetup for Setup<Yes, Yes> {
     type Region = String;
 
     fn region(&self) -> Self::Region {
```
znewman01, line 123: nit: this method name is no longer 100% accurate

src/providers/aws.rs
```diff
@@ -389,6 +389,20 @@ where
...
 
+#[derive(Debug, Clone)]
+struct IpInfo {
+    dns: String,
+    pub_ip: String,
+    priv_ip: String,
+}
+
+#[derive(Debug, Clone)]
+struct TaggedSetup {
```
znewman01, line 404: maybe doc this? not obvious at first glance what "TaggedSetup" means

src/providers/azure.rs
```diff
@@ -459,7 +459,7 @@ impl std::str::FromStr for Region {
             "southafricanorth" => Region::SouthAfricaNorth,
             "uaenorth" => Region::UaeNorth,
             "germanywestcentral" => Region::GermanyWestCentral,
-            u => bail!("Unknown azure region {}", u),
+            u => bail!("Unknown azure region {}. Valid regions: eastus, eastus2, westus, centralus, northcentralus, southcentralus, northeurope, westeurope, eastasia, southeastasia, japaneast, japanwest, australiaeast, australiasoutheast, australiacentral, brazilsouth, southindia, centralindia, westindia, canadacentral, canadaeast, westus2, westcentralus, uksouth, ukwest, koreacentral, koreasouth, francecentral, southafricanorth, uaenorth, germanywestcentral", u),
```
znewman01, line 467: nit: worth pulling out to top of file as const?

src/providers/azure.rs
```diff
@@ -261,13 +267,13 @@ impl super::Launcher for RegionLauncher {
...
         let max_wait = l.max_wait;
-        let vms: Result<Vec<(String, String, _)>, Error> = l.machines
+        let vms: Result<Vec<(String, IpInfo, _)>, Error> = l.machines
             .into_iter()
             .map(|(nickname, desc)| {
                 let vm_name = super::rand_name_sep("vm", "-");
                 debug!(log, "setting up azure instance"; "nickname" => &nickname, "vm_name" => &vm_name);
 
-                let pub_ip = azcmd::create_vm(
+                let ipinfo = azcmd::create_vm(
```
znewman01, line 267: nit: `ip_info` per https://doc.rust-lang.org/1.0.0/style/style/naming/README.html#general-conventions-[rfc-#430]

> In CamelCase, acronyms count as one word: use Uuid rather than UUID. In snake_case, acronyms are lower-cased: is_xid_start.

znewman01 (commented):
Some teeny-tiny things inline.

Overall LGTM!

My high-level design hot take: I actually do like

- the one-entry-point-to-rule-them-all aspect of `TsunamiBuilder`
- one namespace for all my machines

That is, in my head, I want to (pseudo-Rust)

```rust
let tb = ...;
let aws_launcher = ...;
let azure_launcher = ...;
tb.add("server", aws_launcher.make_machine());
tb.add_many("client", 42, aws_launcher.make_machine());
let tsunami = tb.spawn();
tsunami["server"].ssh.command("shutdown");
for client in tsunami["client-{0,1,2}"] { client.ssh.command("ls -l"); }
```

I also have one more nitpick, which is that I find the terminate-resources-in-`Drop` to be a bit unwieldy, especially once there's async stuff involved (the implementation is kind of a hack, and there's [no](https://internals.rust-lang.org/t/asynchronous-destructors/11127) [better](https://boats.gitlab.io/blog/post/poll-drop/) way to do this right now.

If we were to go back to the "do everything with machines in a closure" pattern (which I like!), I think that's pretty unfortunate if I have to do:

```rust
aws_launcher.connect_all(..., |vms| {
   azure_launcher.connect_all(..., |vms| {
      // ...
   });
});
```

both ergonomically and in terms of trying to spawn all the machines at the same time.

(Feel free to ignore for now and address later, or ignore forever!)

akshayknarayan:
Thanks for the comments!

> I also have one more nitpick, which is that I find the terminate-resources-in-Drop to be a bit unwieldy, especially once there's async stuff involved (the implementation is kind of a hack, and there's no better way to do this right now.

Yeah, it definitely sucks, and when writing #31 today I didn't really have a good solution for it. The closure idea (or, in async tsunami, a closure that returns a future) might work, but I agree I dislike the nested calls to `connect_all`.

akshayknarayan:
Closing to fold into #23.

PR #31, opened 2020-04-22
Async Tsunami

akshayknarayan:
See 7815f69.

There are a couple things that I don't like about this, and it would be good to think of ways to get around them.
1. During cleanup, https://docs.rs/tokio/0.2.18/tokio/runtime/struct.Handle.html#method.try_current gives us a `Handle`. `Handle`s can `spawn`, but not `block_on`, so the `impl Drop` on `aws::Launcher` can't block on the cleanup (https://github.com/jonhoo/tsunami/compare/master...akshayknarayan:async-new?expand=1#diff-60699702750aef5a8cdc050b4fce64eaR337).
2. Since ssh is now async, the setup functions, become much more annoying. The approach here has the setup function take a closure which takes a `(&'r ssh::Session, &'r slog::Logger)` and returns a `Pin<Box<dyn Future<Output=...> + 'r>>`. The result is... not pretty: https://github.com/jonhoo/tsunami/compare/master...akshayknarayan:async-new?expand=1#diff-60699702750aef5a8cdc050b4fce64eaR220


jonhoo:
Hmm, the links don't seem to work for me?

For 1, yeah, this is the classic async drop problem. In general, the way you do this is that you provide an `async fn close(self)` that you expect people will call (and document that they should), and then `Drop` is just best-effort by spawning the cleanup future.

For 2 what we really want is async closures, which we don't have yet. Could we make `setup` be generic over `F where F: Fn() -> R, R: Future`?

akshayknarayan:
> For 1, yeah, this is the classic async drop problem. In general, the way you do this is that you provide an async fn close(self) that you expect people will call (and document that they should), and then Drop is just best-effort by spawning the cleanup future.

Yeah, I've added an `fn cleanup()` to the `Launcher` trait.

> For 2 what we really want is async closures, which we don't have yet. Could we make setup be generic over F where F: Fn() -> R, R: Future?

I tried this:
```rust
    pub fn setup<F, R>(
        mut self,
        setup: F,
    ) -> Self
    where F: for<'r> Fn(
                &'r mut ssh::Session,
                &'r slog::Logger,
            ) -> R + 'r
            + Send
            + Sync
            + 'static,
        R: Future<Output = Result<(), Error>> + Send
    {
        self.setup_fn = Some(Arc::new(|ssh, log| Box::pin(setup(ssh, log))));
        self
    }
```
which doesn't work:
```
error[E0261]: use of undeclared lifetime name `'r`
   --> src/providers/aws.rs:228:22
    |
228 |             ) -> R + 'r
    |                      ^^ undeclared lifetime
    |
help: consider introducing lifetime `'r` here
    |
157 | impl<'r> Setup {
    |     ^^^^
help: consider introducing lifetime `'r` here
    |
221 |     pub fn setup<'r, F, R>(
    |                  ^^^
```

I think this is somewhat hopeless because `(Fn(...) -> R + 'r) + Send + ...` vs `(Fn(...) -> R + 'r + Send + Sync + 'static)` is ambiguous.

The current way was the only way I could think of to limit the future returned by the closure by the lifetime of one of the closure arguments.

akshayknarayan:
(see attempt here: https://github.com/akshayknarayan/tsunami/blob/setup-async-closure/src/providers/aws.rs#L221)

jonhoo:
Hmm.. One option here is to introduce a `Setup` trait instead. Then I think you could do it?

akshayknarayan:
Folding into #23.

PR #32, opened 2020-05-20
Move everything to color_eyre + tracing

jonhoo:


jonhoo:
This has been a fairly mechanical transformation. I'm sure I've messed _all sort of things_ up.

@hawkw and @yaahc, if you want to see the end result when someone mostly skims the documentation during a complete, decently sizeable, and fairly quick rewrite from `failure` and `slog` to `color_eyre` and `tracing`, here's that :sweat_smile: If you have spare cycles and want to pattern match on whether I made any obvious oopsies, it'd be much appreciated, but if not, no worries at all. Just figured it might be a useful UX datapoint for what a bigger transition might look like (and the errors people are likely to make — I'm sure I've made several).

jonhoo:
The various logging levels are pretty ad-hoc at the moment, both for events and spans. I'm also sure we're missing a lot of _events_, even though we're decently well covered on spans (mostly due to `#[instrument]`).

src/providers/aws.rs
```diff
@@ -17,13 +17,7 @@
...
 //!     // make the defined-duration instances expire after 1 hour
 //!     l.set_max_instance_duration(1);
-//!     l.spawn(
-//!         vec![(String::from("my machine"), aws::Setup::default())],
-//!         None,
-//!         None,
-//!     )
-//!     .await
-//!     .unwrap();
+//!     l.spawn(vec![(String::from("my machine"), aws::Setup::default())], None).await.unwrap();
```
akshayknarayan, line 20: did this line change? Do you have a different rustfmt than me?

src/lib.rs
```diff
@@ -102,6 +100,19 @@
...
 //! ```
 //!
+//! # Where are the logs?
+//!
+//! This crate uses [`tracing`](https://docs.rs/tracing), which does not log anything by default,
+//! since the crate allows you to [plug and
+//! play](https://docs.rs/tracing/0.1.14/tracing/#in-executables) which "consumer" you want for
+//! your trace points. If you want logging that "just works", you'll want
+//! [`tracing_subscriber::fmt`](https://docs.rs/tracing-subscriber/0.2/tracing_subscriber/fmt/index.html),
+//! which you can instantiate (after adding it to your Cargo.toml) with:
```
hawkw, line 110: Maybe the docs should also mention the following:
 - users of the `log` crate can enable `log` support by depending on `tracing` and enabling the "log" feature flag
 - i _believe_ (though @yaahc can probably confirm this) that `color-eyre` supports `tracing-error` integration as well, so it may be worth suggesting that users use the `tracing-error` crate as well?

src/providers/aws.rs
```diff
@@ -527,24 +535,26 @@ where
         })
     }
 
+    #[instrument(debug)]
```
hawkw: I think you meant
```suggestion
    #[instrument(level = "debug")]
```

src/providers/aws.rs
```diff
@@ -527,24 +535,26 @@ where
...
-    ) -> Pin<Box<dyn Future<Output = Result<HashMap<String, crate::Machine<'l>>, Error>> + Send + 'l>>
-    {
+    ) -> Pin<
+        Box<dyn Future<Output = Result<HashMap<String, crate::Machine<'l>>, Report>> + Send + 'l>,
+    > {
         Box::pin(async move { collect!(self.regions) })
     }
 
-    fn terminate_all(mut self) -> Pin<Box<dyn Future<Output = Result<(), Error>> + Send>> {
+    #[instrument(debug)]
```
hawkw: I think you meant
```suggestion
    #[instrument(level = "debug")]
```

src/providers/aws.rs
```diff
@@ -607,32 +616,38 @@ impl RegionLauncher {
...
+            .wrap_err("failed to make security groups")?
             .make_ssh_key()
-            .await?;
+            .await
+            .wrap_err("failed to make ssh key")?;
 
         Ok(ec2)
     }
 
+    #[instrument(debug, skip(provider))]
```
hawkw: I think you meant
```suggestion
    #[instrument(level = "debug", skip(provider))]
```

src/providers/aws.rs
```diff
@@ -641,59 +656,67 @@ impl RegionLauncher {
...
-            log: Some(log),
         })
     }
 
     /// Region-specific instance setup.
     ///
     /// Make spot instance requests, wait for the instances, and then call the
     /// instance setup functions.
-    pub async fn launch(
+    #[instrument(debug, skip(self, max_instance_duration_hours, max_wait))]
```
hawkw:
```suggestion
    #[instrument(level = "debug", skip(self, max_instance_duration_hours, max_wait))]
```

src/providers/aws.rs
```diff
@@ -641,59 +656,67 @@ impl RegionLauncher {
...
-        self.wait_for_instances(max_wait).await?;
+        self.wait_for_instances(max_wait)
+            .await
+            .wrap_err("failed while waiting for instances to come up")?;
         Ok(())
     }
 
-    async fn make_security_group(mut self, use_open_ports: bool) -> Result<Self, Error> {
-        let log = self.log.as_ref().expect("RegionLauncher uninitialized");
+    #[instrument(trace, skip(self))]
```
hawkw:
```suggestion
    #[instrument(level = "trace", skip(self))]
```

src/providers/aws.rs
```diff
@@ -744,33 +767,33 @@ impl RegionLauncher {
...
-            .context("failed to fill in security group for new machines")?;
+            .wrap_err("failed to fill in security group for new machines")?;
 
         self.security_group_id = group_id;
         Ok(self)
     }
 
-    async fn make_ssh_key(mut self) -> Result<Self, Error> {
-        let log = self.log.as_ref().expect("RegionLauncher uninitialized");
+    #[instrument(trace, skip(self))]
```
hawkw:
```suggestion
    #[instrument(level = "trace", skip(self))]
```

src/providers/aws.rs
```diff
@@ -795,15 +821,20 @@ impl RegionLauncher {
     ///
     /// Will *not* wait for the spot instance requests to complete. To wait, call
     /// [`wait_for_spot_instance_requests`](RegionLauncher::wait_for_spot_instance_requests).
-    async fn make_spot_instance_requests(
+    #[instrument(trace, skip(self, max_duration))]
```
hawkw:
```suggestion
    #[instrument(level = "trace", skip(self, max_duration))]
```

src/providers/aws.rs
```diff
@@ -917,36 +950,31 @@ impl RegionLauncher {
     ///
     /// To wait for the instances to be ready, call
     /// [`wait_for_instances`](RegionLauncher::wait_for_instances).
+    #[instrument(trace, skip(self, max_wait))]
```
hawkw:
```suggestion
    #[instrument(level = "trace", skip(self, max_wait))]
```

src/providers/aws.rs
```diff
@@ -1017,55 +1046,50 @@ impl RegionLauncher {
...
+                eyre::bail!("wait limit reached");
             }
         }
 
         Ok(())
     }
 
     /// Poll AWS until `max_wait` (if not `None`) or the instances are ready to SSH to.
-    async fn wait_for_instances(&mut self, max_wait: Option<time::Duration>) -> Result<(), Error> {
+    #[instrument(trace, skip(self, max_wait))]
```
hawkw:
```suggestion
    #[instrument(level = "trace", skip(self, max_wait))]
```

src/providers/aws.rs
```diff
@@ -1187,36 +1210,40 @@ impl RegionLauncher {
 
     /// Establish SSH connections to the machines. The `Ok` value is a `HashMap` associating the
     /// friendly name for each `Setup` with the corresponding SSH connection.
-    pub async fn connect_all<'l>(&'l self) -> Result<HashMap<String, crate::Machine<'l>>, Error> {
-        let log = self.log.as_ref().unwrap();
+    #[instrument(debug)]
```
hawkw:
```suggestion
    #[instrument(level = "debug")]
```

src/providers/aws.rs
```diff
@@ -1227,53 +1254,53 @@ impl RegionLauncher {
     ///
     /// Additionally deletes ephemeral keys and security groups. Note: it is a known issue that
     /// security groups often will not be deleted, due to timing quirks in the AWS api.
+    #[instrument(debug)]
```
hawkw:
```suggestion
    #[instrument(level = "debug")]
```

src/providers/aws.rs
```diff
@@ -1227,53 +1254,53 @@ impl RegionLauncher {
...
-                    "error" => ?e,
-                )
+                tracing::warn!("failed to clean up temporary security group: {}", e);
             }
         }
 
         if !self.ssh_key_name.trim().is_empty() {
-            trace!(log, "cleaning up temporary keypair"; "name" => self.ssh_key_name.clone());
+            let key_span = tracing::trace_span!("key", name = %self.ssh_key_name);
+            let _guard = key_span.enter();
```
hawkw: It looks like these span entered guards are held across await points, which is probably going to result in malformed traces. You should use `async` blocks and `.instrument` instead. See https://github.com/tokio-rs/tracing/#in-asynchronous-code

src/providers/azure.rs
```diff
@@ -329,71 +332,73 @@ impl RegionLauncher {
 impl super::Launcher for RegionLauncher {
     type MachineDescriptor = Setup;
 
+    #[instrument(debug, skip(self))]
```
hawkw:
```suggestion
    #[instrument(level = "debug", skip(self))]
```

src/providers/aws.rs
```diff
@@ -779,7 +802,10 @@ impl RegionLauncher {
         private_key_path
             .write_all(private_key.as_bytes())
             .context("could not write private key to file")?;
-        debug!(log, "wrote keypair to file"; "filename" => private_key_path.path().display());
+        tracing::trace!(
+            filename = %private_key_path.path().display(),
+            "wrote keypair to file"
+        );
```
akshayknarayan, line 815: I think we should keep this as a `debug!` - often when debugging a script I'll look back at the output to manually ssh to the machine (i.e., copy paste the output of this log statement to `ssh -i <foo> ubuntu@ip`)

src/providers/azure.rs
```diff
@@ -329,71 +332,73 @@ impl RegionLauncher {
...
+                }))
+                .await
+                .into_iter()
+                .collect::<Result<Vec<_>, Report>>()?;
 
             Ok(())
         })
     }
 
+    #[instrument(debug)]
```
hawkw:
```suggestion
    #[instrument(level = "debug")]
```

src/providers/azure.rs
```diff
@@ -413,18 +418,19 @@ impl super::Launcher for RegionLauncher {
...
             .await
             .into_iter()
-            .collect::<Result<HashMap<_, _>, Error>>()
+            .collect::<Result<HashMap<_, _>, Report>>()
         })
     }
 
-    fn terminate_all(self) -> Pin<Box<dyn Future<Output = Result<(), Error>> + Send>> {
-        debug!(self.log.as_ref().unwrap(), "Cleaning up resource group");
+    #[instrument(debug)]
```
hawkw:
```suggestion
    #[instrument(level = "debug")]
```

src/providers/azure.rs
```diff
@@ -557,28 +563,28 @@ impl std::str::FromStr for Region {
...
+    pub(crate) async fn check_az() -> Result<(), Report> {
+        eyre::ensure!(
+            Command::new("az").arg("account").arg("show").status().await.wrap_err("az account show")?.success(), 
             "Azure CLI not found. See https://docs.microsoft.com/en-us/cli/azure/install-azure-cli?view=azure-cli-latest for installation, then run `az login`.",
         );
         Ok(())
     }
 
-    pub(crate) async fn create_resource_group(r: Region, name: &str) -> Result<(), Error> {
+    #[instrument(trace)]
```
hawkw:
```suggestion
    #[instrument(level = "trace")]
```

src/providers/azure.rs
```diff
@@ -589,22 +595,21 @@ mod azcmd {
...
-            bail!("Failed to create resource group {} in region {:?}", name, r)
-        }
+            .await
+            .context("az group create")?;
 
+        eyre::ensure!(out.success(), "failed to create resource group");
         Ok(())
     }
 
+    #[instrument(trace)]
```
hawkw:
```suggestion
    #[instrument(level = "trace")]
```

src/providers/azure.rs
```diff
@@ -631,23 +636,26 @@ mod azcmd {
...
+        eyre::ensure!(vm.powerState == "VM running", "VM power state incorrect");
+        eyre::ensure!(vm.resourceGroup == rg, "VM resource group incorrect");
         Ok(IpInfo {
             public_ip: vm.publicIpAddress,
             private_ip: vm.privateIpAddress,
         })
     }
 
-    pub(crate) async fn open_ports(rg: &str, vm_name: &str) -> Result<(), Error> {
+    #[instrument(trace)]
```
hawkw:
```suggestion
    #[instrument(level = "trace")]
```

src/providers/azure.rs
```diff
@@ -660,32 +668,35 @@ mod azcmd {
...
+            out.status.success(),
+            "failed to open ports: {}",
+            String::from_utf8_lossy(&out.stderr)
+        );
 
         Ok(())
     }
 
-    pub(crate) async fn delete_resource_group(rg: &str) -> Result<(), Error> {
+    #[instrument(trace)]
```
hawkw:
```suggestion
    #[instrument(level = "trace")]
```

src/providers/baremetal.rs
```diff
@@ -43,12 +44,18 @@ impl super::MachineSetup for Setup {
 
 impl Setup {
     /// Create a new instance of Setup.
-    pub fn new(
-        addr: impl std::net::ToSocketAddrs,
+    #[instrument(debug)]
```
hawkw:
```suggestion
    #[instrument(level = "debug")]
```

src/providers/baremetal.rs
```diff
@@ -171,26 +170,27 @@ pub struct Machine {
 impl super::Launcher for Machine {
     type MachineDescriptor = Setup;
 
+    #[instrument(debug, skip(self))]
```
hawkw:
```suggestion
    #[instrument(level = "debug", skip(self))]
```

src/providers/baremetal.rs
```diff
@@ -242,15 +230,14 @@ impl super::Launcher for Machine {
         })
     }
 
+    #[instrument(debug)]
```
hawkw:
```suggestion
    #[instrument(level = "debug")]
```

hawkw (commented):
This is awesome! I noticed a couple issues you'll probably want to fix, but overall this is super nice to see :)

akshayknarayan (approved):
Don't we have to call `tracing_subscriber::fmt::init()` in the examples? Other than that, looks good :)

yaahc:
Hype! Checking it out rn.

src/lib.rs
yaahc: >  that `color-eyre` supports `tracing-error` integration

This is correct, also it is enabled by default, so if you don't explicitly disable the `capture-spantrace` feature in `color-eyre` they will already have `tracing-error` in their dependency tree. Definitely worth recommending that they setup an error layer.

src/providers/aws.rs
jonhoo: Ah, that's really sad. I guess that means a lot more code indentation :'(

jonhoo: That extra block also means an extra row in `eyre` backtraces I think, since there's now one more `?` (for the extra `async { ... }.await?`).

jonhoo: Ended up being this: https://github.com/jonhoo/tsunami/pull/32/commits/084205c6991a6d86a473221cde5a7269da010eb1

jonhoo: @yaahc tagging you specifically here for the observation about eyre backtraces with the extra `?` above.

src/providers/aws.rs
jonhoo: Yup, it changed! One fewer `None` (for the logger).

jonhoo:
@akshayknarayan I mean, we don't have to if we don't care about their output :p Whether logging is enabled or not is sort of immaterial to what the examples try to show I think?

src/providers/azure.rs
```diff
@@ -557,28 +579,28 @@ impl std::str::FromStr for Region {
             r if r == Region::SouthAfricaNorth.as_ref() => Region::SouthAfricaNorth,
             r if r == Region::UaeNorth.as_ref() => Region::UaeNorth,
             r if r == Region::GermanyWestCentral.as_ref() => Region::GermanyWestCentral,
-            u => bail!("Unknown azure region {}. Valid regions: eastus, eastus2, westus, centralus, northcentralus, southcentralus, northeurope, westeurope, eastasia, southeastasia, japaneast, japanwest, australiaeast, australiasoutheast, australiacentral, brazilsouth, southindia, centralindia, westindia, canadacentral, canadaeast, westus2, westcentralus, uksouth, ukwest, koreacentral, koreasouth, francecentral, southafricanorth, uaenorth, germanywestcentral", u),
+            u => eyre::bail!("Unknown azure region {}. Valid regions: eastus, eastus2, westus, centralus, northcentralus, southcentralus, northeurope, westeurope, eastasia, southeastasia, japaneast, japanwest, australiaeast, australiasoutheast, australiacentral, brazilsouth, southindia, centralindia, westindia, canadacentral, canadaeast, westus2, westcentralus, uksouth, ukwest, koreacentral, koreasouth, francecentral, southafricanorth, uaenorth, germanywestcentral", u),
```
yaahc: I would move the `Valid regions:` bit here into the help text in `color_eyre` via the `Help` trait. I should probably add an `Info` help type for this, right now color-eyre only has categories for suggestions, warnings, and notes.

src/lib.rs
jonhoo: For `tracing-error`, I actually added a short bit on that just after you commented this :p

src/providers/azure.rs
```diff
@@ -631,23 +652,26 @@ mod azcmd {
...
+            .wrap_err("az vm create")?;
 
-        if !out.status.success() {
-            return Err(format_err!("Failed to create vm {}", name))
-                .context(String::from_utf8(out.stderr).unwrap())?;
-        }
+        eyre::ensure!(
+            out.status.success(),
+            "failed to create vm: {}",
+            String::from_utf8_lossy(&out.stderr)
```
yaahc, line 661: Now I want to add support for custom sections to `color-eyre` I think it would be nice if it could print a report like:

```
Error:
     0: ...

Stderr:
     ....

---- SPANTRACE ----
....
```

Where you could make it with an API like this

```rust
eyre!("failed to create vm").add_section(Section{ title: "Stderr", body: String::from_utf8_lossy(&out.stderr)})
```

src/providers/aws.rs
jonhoo: Going to leave this unresolved just so the discussion doesn't disappear entirely.

src/providers/azure.rs
jonhoo: [Like so](https://github.com/jonhoo/tsunami/pull/32/commits/1e0ea8b7815ffb20e41657819445d43b7412437b) I think?

src/providers/baremetal.rs
```diff
@@ -43,12 +44,18 @@ impl super::MachineSetup for Setup {
...
+        addr: A,
         username: Option<String>,
-    ) -> Result<Self, Error> {
-        let username: Result<String, Error> = username.map(Ok).unwrap_or_else(|| {
-            let user = String::from_utf8(std::process::Command::new("whoami").output()?.stdout)?;
+    ) -> Result<Self, Report> {
+        let username: Result<String, Report> = username.map(Ok).unwrap_or_else(|| {
+            let stdout = std::process::Command::new("whoami")
+                .output()
+                .wrap_err("whoami to determine local user")?
```
yaahc: this looks like code smell to me, I'd either change this message to represent an actual error or move it into a span that is captured by `tracing-error` when the report is created. Last time I ran into something similar I ended up making an extension trait for `Command` that would create a span which records the cmd as debug.

src/providers/azure.rs
jonhoo: Ooooh, yes, that'd be awesome! That way I could include both stderr, the exit status code, and other stuff!

src/providers/baremetal.rs
```diff
@@ -171,26 +170,27 @@ pub struct Machine {
...
+                    name = %discarded_name,
+                    addr = %discarded_setup.addr[0],
+                    "Discarding duplicate connections to same machine",
                 );
             }
 
-            let addr = try_addrs(&mut setup, &log, l.max_wait).await?;
+            let addr = try_addrs(&mut setup, l.max_wait)
+                .await
+                .wrap_err("probing baremetal addresses")?;
```
yaahc: similar to above, this seems like something that belongs in a span or needs to be rewritten to represent an error.

akshayknarayan:
> @akshayknarayan I mean, we don't have to if we don't care about their output :p Whether logging is enabled or not is sort of immaterial to what the examples try to show I think?

Fair enough, but at least https://github.com/jonhoo/tsunami/blob/master/examples/launch.rs#L50 and https://github.com/jonhoo/tsunami/blob/master/examples/test1.rs#L66 should be updated no?

src/providers/baremetal.rs
jonhoo: Yeah, this one is a little tricky. I don't disagree with you in principle, but this isn't a common pattern in this particular application, and the only thing executed in that span would also be this command. A span or a command wrapper seems like overkill for this one case?

src/providers/azure.rs
yaahc: precisely

src/providers/baremetal.rs
yaahc: Not sure, what do you mean by overkill?

This is what I'm imagining.

```rust
trait Output {
    fn output2(&mut self) -> Result<String, Report>;
}

impl Output for Command {
    #[instrument]
    fn output2(&mut self) -> Result<String, Report> {
        let output = self.output()?;

        let stdout = String::from_utf8_lossy(&output.stdout);

        if !output.status.success() {
            let stderr = String::from_utf8_lossy(&output.stderr);
            eyre!("cmd exited with non-zero status code")
                .add_section(Section { title: "stdout:", body: stdout })
                .add_section(Section { title: "stderr:", body: stderr})
        } else {
            Ok(stdout.into())
        }
    }
}
```

I lean towards it being worth it just because it makes debugging errors from subshells WAY nicer.

src/lib.rs
```diff
@@ -113,10 +113,13 @@
...
 //!
-//! And then run your application with, for example, `RUST_LOG=info` to get logs. If you also want
-//! automatic logging of errors, take a look at the documentation for
-//! [`color-eyre`](https://docs.rs/color_eyre/), which includes information about how to integrate
-//! with [`tracing-error`](https://docs.rs/tracing-error).
+//! And then run your application with, for example, `RUST_LOG=info` to get logs. If you're using
+//! the `log` crate, you can instead just add a dependency on `tracing` with the `log` feature
+//! enabled, and things should just "magically" work.
+//!
+//! If you also want automatic logging of errors (which I think you do), take a look at the
```
yaahc: tracing-error doesn't do automatic logging of errors. It just provides `SpanTrace` which gives automatic capture of spans with errors which it then prints in the report.

src/providers/baremetal.rs
jonhoo: Only tangentially related, but the code this ends up calling (`try_addrs`) actually accumulates _multiple_ errors, and I'm not sure how to nicely collapse that into one error. What I'm now doing is
https://github.com/jonhoo/tsunami/blob/1e0dadc760890e90d9596cfb575dae9b73ccd298/src/providers/baremetal.rs#L170-L177

but that feels pretty dirty... Ideas?

src/lib.rs
jonhoo: Ah, sorry, what I meant to imply here (which, maybe I'm wrong?) was that `color_eyre` will create those `SpanTrace` for you whenever it propagates an error? But maybe I just completely made that up?

jonhoo:
@akshayknarayan Ah, yes, seems I forgot to even try to compile the examples :sweat_smile:

src/providers/baremetal.rs
yaahc: aah, yes, I definitely have thoughts about this. The error trait doesn't really allow for representing errors as anything other than a singly linked list.

Using `wrap_err` here is probably something you definitely don't want to do, because it will use the `Display` method for the errors which will only show the first error message in that report's chain of errors.

Ideally, I'd say don't try to represent errors past this point as errors anymore. Just stick with the language of error reports and return a Vec<Report> if possible.

That said there may be a way to construct an error with multiple source errors in the near future. I have an [RFC](https://github.com/yaahc/rfcs/blob/master/text/0000-dyn-error-generic-member-access.md#example-use-cases-this-enables) in progress that actually makes this possible. Specifically, check out the 3rd bullet in list i linked to about representing source errors as a tree of errors. If that RFC ever merges I'll update eyre and company to be able to wrap single errors or sets of errors, and correctly traverse the tree of errors when reporting.

src/lib.rs
yaahc: you're not wrong, this is what `color_eyre` does, though I would say it does the capture when its converted, not when its propogated, though that's often the same thing. I think the distinction is important because I'm pushing to add error return traces which would be specifically tied to propogation.

src/providers/baremetal.rs
jonhoo: Hmm, a `Vec<Report>` seems pretty unwieldy though. Unless I change all error types in the application to `Vec<Report>` to cater to this specific cast. I think defaulting to just displaying the first is actually perfectly fine in this _particular_ case. The error here is if a particular host has multiple IP addresses, and we fail to connect to _all_ of them. Reporting the error just for the first IP address in the common case seems fine to me. Thanks for the RFC link though, will definitely be watching that!

src/providers/baremetal.rs
jonhoo: I guess what I meant is that in this crate, the only place that extension trait would be used would be here for `whoami` — a command that is _very_ unlikely to fail in the first place. Just reporting that "whoami failed" I think is enough context for this particular case. In a program that actually executes more commands though, completely agree this would be useful.

src/lib.rs
jonhoo: If I'm understanding you correctly, I then still think this text is accurate? All it's saying is that _if_ you follow the `color_eyre` example, you get automatic logging of errors. Which it then sounds like is accurate. How do you think it should be rephrased?

src/providers/baremetal.rs
yaahc: oh, I definitely agree with that.

src/lib.rs
yaahc: The confusion I'm worried about is people thinking that converting to a `color_eyre::Report` somehow triggers a tracing event which literally logs the error the moment it is constructed. Like, if you drop a report immediately after making it nothing gets logged, and I don't want people thinking that it would log.

I would probably just tweek it to "automatically include tracing spans with errors" or maybe "log tracing spans with your errors automatically" or something else along those lines. Just make it clear that it's not providing automatic _logging_.

jonhoo: How's https://github.com/jonhoo/tsunami/pull/32/commits/80b29d0afd59935abb90f78c53310a0a7a5da947 ?

yaahc (approved):

src/lib.rs
yaahc: looks perfect

jonhoo:
@hawkw @yaahc Thank you so much for giving such an in-depth review! Was not expecting that :heart:

PR #33, opened 2020-05-21
Retry loop to delete aws security groups

### LINKED ISSUE #2: Bring back cleanup code
Code has been written for cleaning up the security group and keypair generated by `tsunami` for every run, but it is currently commented out:

https://github.com/jonhoo/tsunami/blob/379acb77aa8b71a4c92d91342ce76d8a9572e26e/src/lib.rs#L445-L458

This is because it turns out that EC2 won't let you delete resources that are associated with instances that have not yet been terminated. We ran into this during the live-coding session @
https://youtu.be/66INYb73yXo?t=5716. The problem is specifically that although we send a termination request for all the instances, they may not yet have terminated by the time we try to do cleanup (and the deletion of the security group thus fails). The way around this would be to keep checking on all the instances until they have *all* been terminated, and only *then* do cleanup.

akshayknarayan:
In my tests, it seems to take about 15 seconds of retrying, so the retry loop is 5 sec long. We could do fancier things like exponential backoff, but this is good enough for now I think. 

Fixes #2

```
May 21 11:37:30.811 TRACE terminate_all{self=Launcher { max_instance_duration_hours: 1, use_open_ports: false, regions: {"us-east-1": RegionLauncher { region: UsEast1, availability_zone: Any, security_group_id: "s
g-0cf3db4d6b3582264", ssh_key_name: "tsunami_key_LgHBYrBD5A", private_key_path: Some(NamedTempFile("/tmp/.tmpRMFqkL")), outstanding_spot_request_ids: {}, instances: {"i-02bf10b045dcea105": TaggedSetup { name: "my
machine", setup: Setup { region: UsEast1, availability_zone: Any, instance_type: "t3.small", ami: "ami-085925f297f89fce1", username: "ubuntu" }, ip_info: Some(IpInfo { public_dns: "ec2-54-234-116-36.compute-1.amaz
onaws.com", public_ip: "54.234.116.36", private_ip: "172.31.3.193" }) }} }} }}:region{region=us-east-1}:terminate_all{self=RegionLauncher { region: UsEast1, availability_zone: Any, security_group_id: "sg-0cf3db4d6
b3582264", ssh_key_name: "tsunami_key_LgHBYrBD5A", private_key_path: Some(NamedTempFile("/tmp/.tmpRMFqkL")), outstanding_spot_request_ids: {}, instances: {"i-02bf10b045dcea105": TaggedSetup { name: "my machine", s
etup: Setup { region: UsEast1, availability_zone: Any, instance_type: "t3.small", ami: "ami-085925f297f89fce1", username: "ubuntu" }, ip_info: Some(IpInfo { public_dns: "ec2-54-234-116-36.compute-1.amazonaws.com",
 public_ip: "54.234.116.36", private_ip: "172.31.3.193" }) }} }}:security_group{id=sg-0cf3db4d6b3582264}: tsunami::providers::aws: removing security group.
May 21 11:37:31.254 TRACE terminate_all{self=Launcher { max_instance_duration_hours: 1, use_open_ports: false, regions: {"us-east-1": RegionLauncher { region: UsEast1, availability_zone: Any, security_group_id: "s
g-0cf3db4d6b3582264", ssh_key_name: "tsunami_key_LgHBYrBD5A", private_key_path: Some(NamedTempFile("/tmp/.tmpRMFqkL")), outstanding_spot_request_ids: {}, instances: {"i-02bf10b045dcea105": TaggedSetup { name: "my
machine", setup: Setup { region: UsEast1, availability_zone: Any, instance_type: "t3.small", ami: "ami-085925f297f89fce1", username: "ubuntu" }, ip_info: Some(IpInfo { public_dns: "ec2-54-234-116-36.compute-1.amaz
onaws.com", public_ip: "54.234.116.36", private_ip: "172.31.3.193" }) }} }} }}:region{region=us-east-1}:terminate_all{self=RegionLauncher { region: UsEast1, availability_zone: Any, security_group_id: "sg-0cf3db4d6
b3582264", ssh_key_name: "tsunami_key_LgHBYrBD5A", private_key_path: Some(NamedTempFile("/tmp/.tmpRMFqkL")), outstanding_spot_request_ids: {}, instances: {"i-02bf10b045dcea105": TaggedSetup { name: "my machine", s
etup: Setup { region: UsEast1, availability_zone: Any, instance_type: "t3.small", ami: "ami-085925f297f89fce1", username: "ubuntu" }, ip_info: Some(IpInfo { public_dns: "ec2-54-234-116-36.compute-1.amazonaws.com",
 public_ip: "54.234.116.36", private_ip: "172.31.3.193" }) }} }}:security_group{id=sg-0cf3db4d6b3582264}: tsunami::providers::aws: failed to clean up temporary security group: Request ID: None Body: <?xml version=
"1.0" encoding="UTF-8"?>
<Response><Errors><Error><Code>DependencyViolation</Code><Message>resource sg-0cf3db4d6b3582264 has a dependent object</Message></Error></Errors><RequestID>59046c08-1c08-4a84-a5c5-548cbbb259cc</RequestID></Respons
e>. retrying.
May 21 11:37:36.873 TRACE terminate_all{self=Launcher { max_instance_duration_hours: 1, use_open_ports: false, regions: {"us-east-1": RegionLauncher { region: UsEast1, availability_zone: Any, security_group_id: "s
g-0cf3db4d6b3582264", ssh_key_name: "tsunami_key_LgHBYrBD5A", private_key_path: Some(NamedTempFile("/tmp/.tmpRMFqkL")), outstanding_spot_request_ids: {}, instances: {"i-02bf10b045dcea105": TaggedSetup { name: "my
machine", setup: Setup { region: UsEast1, availability_zone: Any, instance_type: "t3.small", ami: "ami-085925f297f89fce1", username: "ubuntu" }, ip_info: Some(IpInfo { public_dns: "ec2-54-234-116-36.compute-1.amaz
onaws.com", public_ip: "54.234.116.36", private_ip: "172.31.3.193" }) }} }} }}:region{region=us-east-1}:terminate_all{self=RegionLauncher { region: UsEast1, availability_zone: Any, security_group_id: "sg-0cf3db4d6
b3582264", ssh_key_name: "tsunami_key_LgHBYrBD5A", private_key_path: Some(NamedTempFile("/tmp/.tmpRMFqkL")), outstanding_spot_request_ids: {}, instances: {"i-02bf10b045dcea105": TaggedSetup { name: "my machine", s
etup: Setup { region: UsEast1, availability_zone: Any, instance_type: "t3.small", ami: "ami-085925f297f89fce1", username: "ubuntu" }, ip_info: Some(IpInfo { public_dns: "ec2-54-234-116-36.compute-1.amazonaws.com",
 public_ip: "54.234.116.36", private_ip: "172.31.3.193" }) }} }}:security_group{id=sg-0cf3db4d6b3582264}: tsunami::providers::aws: failed to clean up temporary security group: Request ID: None Body: <?xml version=
"1.0" encoding="UTF-8"?>
<Response><Errors><Error><Code>DependencyViolation</Code><Message>resource sg-0cf3db4d6b3582264 has a dependent object</Message></Error></Errors><RequestID>cebaa992-a8ae-4ad7-a836-95d24393fecd</RequestID></Respons
e>. retrying.
May 21 11:37:42.397 TRACE terminate_all{self=Launcher { max_instance_duration_hours: 1, use_open_ports: false, regions: {"us-east-1": RegionLauncher { region: UsEast1, availability_zone: Any, security_group_id: "s
g-0cf3db4d6b3582264", ssh_key_name: "tsunami_key_LgHBYrBD5A", private_key_path: Some(NamedTempFile("/tmp/.tmpRMFqkL")), outstanding_spot_request_ids: {}, instances: {"i-02bf10b045dcea105": TaggedSetup { name: "my
machine", setup: Setup { region: UsEast1, availability_zone: Any, instance_type: "t3.small", ami: "ami-085925f297f89fce1", username: "ubuntu" }, ip_info: Some(IpInfo { public_dns: "ec2-54-234-116-36.compute-1.amaz
onaws.com", public_ip: "54.234.116.36", private_ip: "172.31.3.193" }) }} }} }}:region{region=us-east-1}:terminate_all{self=RegionLauncher { region: UsEast1, availability_zone: Any, security_group_id: "sg-0cf3db4d6
b3582264", ssh_key_name: "tsunami_key_LgHBYrBD5A", private_key_path: Some(NamedTempFile("/tmp/.tmpRMFqkL")), outstanding_spot_request_ids: {}, instances: {"i-02bf10b045dcea105": TaggedSetup { name: "my machine", s
etup: Setup { region: UsEast1, availability_zone: Any, instance_type: "t3.small", ami: "ami-085925f297f89fce1", username: "ubuntu" }, ip_info: Some(IpInfo { public_dns: "ec2-54-234-116-36.compute-1.amazonaws.com",
 public_ip: "54.234.116.36", private_ip: "172.31.3.193" }) }} }}:security_group{id=sg-0cf3db4d6b3582264}: tsunami::providers::aws: failed to clean up temporary security group: Request ID: None Body: <?xml version=
"1.0" encoding="UTF-8"?>
<Response><Errors><Error><Code>DependencyViolation</Code><Message>resource sg-0cf3db4d6b3582264 has a dependent object</Message></Error></Errors><RequestID>483036cc-69a0-43b8-b3f1-9b3a3c67a965</RequestID></Respons
e>. retrying.
May 21 11:37:47.884 TRACE terminate_all{self=Launcher { max_instance_duration_hours: 1, use_open_ports: false, regions: {"us-east-1": RegionLauncher { region: UsEast1, availability_zone: Any, security_group_id: "s
g-0cf3db4d6b3582264", ssh_key_name: "tsunami_key_LgHBYrBD5A", private_key_path: Some(NamedTempFile("/tmp/.tmpRMFqkL")), outstanding_spot_request_ids: {}, instances: {"i-02bf10b045dcea105": TaggedSetup { name: "my
machine", setup: Setup { region: UsEast1, availability_zone: Any, instance_type: "t3.small", ami: "ami-085925f297f89fce1", username: "ubuntu" }, ip_info: Some(IpInfo { public_dns: "ec2-54-234-116-36.compute-1.amaz
onaws.com", public_ip: "54.234.116.36", private_ip: "172.31.3.193" }) }} }} }}:region{region=us-east-1}:terminate_all{self=RegionLauncher { region: UsEast1, availability_zone: Any, security_group_id: "sg-0cf3db4d6
b3582264", ssh_key_name: "tsunami_key_LgHBYrBD5A", private_key_path: Some(NamedTempFile("/tmp/.tmpRMFqkL")), outstanding_spot_request_ids: {}, instances: {"i-02bf10b045dcea105": TaggedSetup { name: "my machine", s
etup: Setup { region: UsEast1, availability_zone: Any, instance_type: "t3.small", ami: "ami-085925f297f89fce1", username: "ubuntu" }, ip_info: Some(IpInfo { public_dns: "ec2-54-234-116-36.compute-1.amazonaws.com",
 public_ip: "54.234.116.36", private_ip: "172.31.3.193" }) }} }}:security_group{id=sg-0cf3db4d6b3582264}: tsunami::providers::aws: cleaned up temporary security group
test providers::aws::test::make_machine_and_ssh_setupfn ... ok
```

### DIFF
--- a/src/providers/aws.rs
+++ b/src/providers/aws.rs
@@ -1345,8 +1345,10 @@ impl RegionLauncher {
                                 tracing::trace!("instances not yet shut down -- retrying");
                                 tokio::time::delay_for(tokio::time::Duration::from_secs(5)).await;
                             } else {
-                                Err(eyre!(err.to_string()))
-                                    .wrap_err("failed to clean up temporary security group")?;
+                                Err(Report::new(RusotoError::<
+                                    rusoto_ec2::DeleteSecurityGroupError,
+                                >::Unknown(r)))
+                                .wrap_err("failed to clean up temporary security group")?;
                                 unreachable!();
                             }
                         }
--- a/src/providers/aws.rs
+++ b/src/providers/aws.rs
@@ -1302,22 +1302,39 @@ impl RegionLauncher {
             .await;
         }
 
+        use rusoto_core::RusotoError;
         if !self.security_group_id.trim().is_empty() {
-            let group_span = tracing::trace_span!("security_group", id = %self.security_group_id);
+            let group_span =
+                tracing::trace_span!("removing security group", id = %self.security_group_id);
             async {
                 tracing::trace!("removing security group.");
                 // clean up security groups and keys
+                let start = tokio::time::Instant::now();
                 loop {
+                    if start.elapsed() > tokio::time::Duration::from_secs(60) {
+                        tracing::warn!(
+                            "failed to clean up temporary security group after 60 seconds."
+                        );
+                    }
+
                     let mut req = rusoto_ec2::DeleteSecurityGroupRequest::default();
                     req.group_id = Some(self.security_group_id.clone());
-                    if let Err(e) = client.delete_security_group(req).await {
-                        tracing::trace!(
-                            "failed to clean up temporary security group: {}. retrying.",
-                            e,
-                        );
-                        tokio::time::delay_for(tokio::time::Duration::from_secs(5)).await;
-                    } else {
-                        break;
+                    match client.delete_security_group(req).await {
+                        Ok(_) => break,
+                        Err(RusotoError::Unknown(r)) => {
+                            let err = r.body_as_str();
+                            if err.contains("<Code>DependencyViolation</Code>") {
+                                tracing::trace!("instances not yet shut down -- retrying");
+                                tokio::time::delay_for(tokio::time::Duration::from_secs(5)).await;
+                            } else {
+                                tracing::warn!("failed: {}", err);
+                                break;
+                            }
+                        }
+                        Err(e) => {
+                            tracing::warn!("failed: {}", e);
+                            break;
+                        }
                     }
                 }
 
--- a/src/providers/aws.rs
+++ b/src/providers/aws.rs
@@ -556,7 +556,7 @@ where
 
                 futures_util::future::join_all(self.regions.drain().map(|(region, mut rl)| {
                     let region_span = tracing::debug_span!("region", %region);
-                    async move { rl.shutdown().await }.instrument(region_span)
+                    async move { rl.terminate_all().await }.instrument(region_span)
                 }))
                 .await;
                 Ok(())
@@ -1269,7 +1269,7 @@ impl RegionLauncher {
     /// Additionally deletes ephemeral keys and security groups. Note: it is a known issue that
     /// security groups often will not be deleted, due to timing quirks in the AWS api.
     #[instrument(level = "debug")]
-    pub async fn shutdown(&mut self) {
+    pub async fn terminate_all(&mut self) {
         let client = self.client.as_ref().unwrap();
         // terminate instances
         if !self.instances.is_empty() {
@@ -1473,10 +1473,10 @@ mod test {
                     .await?;
 
             if let Err(e) = do_multi_instance_spot_request(&mut ec2).await {
-                ec2.shutdown().await;
+                ec2.terminate_all().await;
                 panic!(e);
             } else {
-                ec2.shutdown().await;
+                ec2.terminate_all().await;
             }
 
             Ok(())
--- a/src/providers/aws.rs
+++ b/src/providers/aws.rs
@@ -554,14 +554,19 @@ where
                     return Ok(());
                 }
 
-                futures_util::future::join_all(self.regions.drain().map(|(region, mut rl)| {
-                    let region_span = tracing::debug_span!("region", %region);
-                    // this throws away a possible error from terminate_all,
-                    // but that is ok.
-                    async move { rl.terminate_all().await }.instrument(region_span)
-                }))
-                .await;
-                Ok(())
+                let res =
+                    futures_util::future::join_all(self.regions.drain().map(|(region, mut rl)| {
+                        let region_span = tracing::debug_span!("region", %region);
+                        async move { rl.terminate_all().await }.instrument(region_span)
+                    }))
+                    .await;
+                res.into_iter().fold(Ok(()), |acc, x| match acc {
+                    Ok(_) => x,
+                    Err(a) => match x {
+                        Ok(_) => Err(a),
+                        Err(e) => Err(a.wrap_err(e)),
+                    },
+                })
             }
             .in_current_span(),
         )
@@ -1331,10 +1336,6 @@ impl RegionLauncher {
                 let start = tokio::time::Instant::now();
                 loop {
                     if start.elapsed() > tokio::time::Duration::from_secs(60) {
-                        tracing::warn!(
-                            "failed to clean up temporary security group after 60 seconds."
-                        );
-
                         Err(Report::msg(
                             "failed to clean up temporary security group after 60 seconds.",
                         ))?;
@@ -1351,13 +1352,12 @@ impl RegionLauncher {
                                 tracing::trace!("instances not yet shut down -- retrying");
                                 tokio::time::delay_for(tokio::time::Duration::from_secs(5)).await;
                             } else {
-                                Err(Report::msg(err.to_string())
-                                    .wrap_err("failed to clean up temporary security group"))?;
+                                Err(eyre!(err.to_string()))
+                                    .wrap_err("failed to clean up temporary security group")?;
                                 unreachable!();
                             }
                         }
                         Err(e) => {
-                            tracing::warn!("failed: {}", e);
                             Err(Report::new(e)
                                 .wrap_err("failed to clean up temporary security group"))?;
                             unreachable!();
--- a/src/providers/aws.rs
+++ b/src/providers/aws.rs
@@ -1313,7 +1313,7 @@ impl RegionLauncher {
                     //    block will get skipped, so sg will get cleaned up below.
                     // 2. there were instances, but we couldn't terminate them. Then, the sg will
                     //    still be attached to them, so there's no point trying to delete it.
-                    Err(e).context("failed to terminate tsunami instances")?;
+                    Err(e).wrap_err("failed to terminate tsunami instances")?;
                     unreachable!();
                 }
             }
@@ -1323,7 +1323,7 @@ impl RegionLauncher {
         if !self.security_group_id.trim().is_empty() {
             let group_span =
                 tracing::trace_span!("removing security group", id = %self.security_group_id);
-            let r = async {
+            async {
                 tracing::trace!("removing security group.");
                 // clean up security groups and keys
                 let start = tokio::time::Instant::now();
@@ -1362,9 +1362,7 @@ impl RegionLauncher {
                 Ok::<_, Report>(())
             }
             .instrument(group_span)
-            .await;
-
-            r?;
+            .await?;
         }
 
         Ok(())
--- a/src/providers/aws.rs
+++ b/src/providers/aws.rs
@@ -1405,6 +1405,7 @@ mod test {
     #[ignore]
     fn make_machine_and_ssh_setupfn() {
         use crate::providers::Launcher;
+        tracing_subscriber::fmt::init();
         let mut rt = tokio::runtime::Runtime::new().unwrap();
         let mut l = super::Launcher::default();
         // make the defined-duration instances expire after 1 hour
@@ -1429,8 +1430,11 @@ mod test {
         let ec2 = RegionLauncher::connect(region, super::AvailabilityZoneSpec::Any, provider)?;
         rt.block_on(async {
             let mut ec2 = ec2.make_ssh_key().await?;
-            println!("==> key name: {}", ec2.ssh_key_name);
-            println!("==> key path: {:?}", ec2.private_key_path);
+            tracing::debug!(
+                name = %ec2.ssh_key_name,
+                path = %ec2.private_key_path.as_ref().unwrap().path().display(),
+                "made key"
+            );
             assert!(!ec2.ssh_key_name.is_empty());
             assert!(ec2.private_key_path.as_ref().unwrap().path().exists());
 
--- a/src/providers/aws.rs
+++ b/src/providers/aws.rs
@@ -1270,12 +1270,31 @@ impl RegionLauncher {
 
     /// Terminate all running instances.
     ///
-    /// Additionally deletes ephemeral keys and security groups. Note: it is a known issue that
-    /// security groups often will not be deleted, due to timing quirks in the AWS api.
+    /// Additionally deletes ephemeral keys and security groups. Sometimes, this deletion can fail
+    /// for various reasons. This method deletes things in this order:
+    /// 1. Try to delete the key pair, but emit a log message and continue if it fails.
+    /// 2. Try to terminate the instances, and short-circuits to return the error if it fails.
+    /// 3. Try to delete the security group. This can fail as the security groups are still
+    ///    "attached" to the instances we just terminated in step 2. So, we retry for 60
+    ///    seconds before giving up and returning an error.
     #[instrument(level = "debug")]
     pub async fn terminate_all(&mut self) -> Result<(), Report> {
-        let mut res = Ok(());
         let client = self.client.as_ref().unwrap();
+
+        if !self.ssh_key_name.trim().is_empty() {
+            let key_span = tracing::trace_span!("key", name = %self.ssh_key_name);
+            async {
+                tracing::trace!("removing keypair");
+                let mut req = rusoto_ec2::DeleteKeyPairRequest::default();
+                req.key_name = self.ssh_key_name.clone();
+                if let Err(e) = client.delete_key_pair(req).await {
+                    tracing::warn!("failed to clean up temporary SSH key: {}", e);
+                }
+            }
+            .instrument(key_span)
+            .await;
+        }
+
         // terminate instances
         if !self.instances.is_empty() {
             tracing::info!("terminating instances");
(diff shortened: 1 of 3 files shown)


src/providers/aws.rs
```diff
@@ -1321,6 +1304,31 @@ impl RegionLauncher {
             .instrument(key_span)
             .await;
         }
+
+        if !self.security_group_id.trim().is_empty() {
+            let group_span = tracing::trace_span!("security_group", id = %self.security_group_id);
+            async {
+                tracing::trace!("removing security group.");
+                // clean up security groups and keys
+                loop {
```
jonhoo, line 1330: We probably want a limit on how long we retry this for. Like a minute or something?

src/providers/aws.rs
```diff
@@ -1321,6 +1304,31 @@ impl RegionLauncher {
...
+                    let mut req = rusoto_ec2::DeleteSecurityGroupRequest::default();
+                    req.group_id = Some(self.security_group_id.clone());
+                    if let Err(e) = client.delete_security_group(req).in_current_span().await {
```
jonhoo: The `in_current_span` here actually isn't necessary. Just pushed a commit to `master` that removes this and some other extraneous ones.

src/providers/aws.rs
```diff
@@ -1321,6 +1304,31 @@ impl RegionLauncher {
...
+                        tracing::trace!(
```
jonhoo: I think here we want to check _what_ error we got. If we get an error _other_ than "busy" or "in use" or whatever it is, then we should propagate that error. The error match against is:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<Response><Errors><Error><Code>DependencyViolation</Code><Message>resource sg-0c3ddd8f4d8bfe33c has a dependent object</Message></Error></Errors><RequestID>196925f6-b9b9-45ed-b4c6-d07798afca37</RequestID></Response>
```
`DependencyViolation` under `<Code>` seems like what we want to look for.

src/providers/aws.rs
```diff
@@ -1321,6 +1304,31 @@ impl RegionLauncher {
...
+                            "failed to clean up temporary security group: {}. retrying.",
```
jonhoo: This should probably just say "instances not yet shut down -- retrying".

src/providers/aws.rs
jonhoo: Alternatively, instead of retrying the delete, we can just keep checking the state of all the instances until all of them are in the `terminated` state. And only then try the delete, which should then succeed immediately. That would mean we don't have to match on the error.

src/providers/aws.rs
```diff
@@ -1302,22 +1302,39 @@ impl RegionLauncher {
...
-                        break;
+                    match client.delete_security_group(req).await {
+                        Ok(_) => break,
+                        Err(RusotoError::Unknown(r)) => {
+                            let err = r.body_as_str();
+                            if err.contains("<Code>DependencyViolation</Code>") {
+                                tracing::trace!("instances not yet shut down -- retrying");
+                                tokio::time::delay_for(tokio::time::Duration::from_secs(5)).await;
+                            } else {
+                                tracing::warn!("failed: {}", err);
```
jonhoo: Here I think we actually want to return an error.

src/providers/aws.rs
```diff
@@ -1302,22 +1302,39 @@ impl RegionLauncher {
...
+                                break;
+                            }
+                        }
+                        Err(e) => {
+                            tracing::warn!("failed: {}", e);
```
jonhoo: And same here.

src/providers/aws.rs
akshayknarayan: Also the timeout case above this needs to be addressed. 
So actually this method currently doesn't return `Result<(), Error>`, it returns `()`. Should we make it return `Result`, in case of timeout? I think this error is not actually fatal (since obviously we have been dealing with failed deletions all this time), and might be fine to keep as a warning (vs users who might be using `?`)

jonhoo: Hmm, it's a good question... I _feel_ like the right thing to do is to have `terminate_all` also return a `Result`. The user is free to ignore errors from `terminate_all` if they wish, but it seems to me like the right thing to do is to at least _report_ that there was an error. That way we can also start to properly return errors for, say, when terminating the instances fails!

src/providers/aws.rs
```diff
@@ -556,6 +556,8 @@ where
 
                 futures_util::future::join_all(self.regions.drain().map(|(region, mut rl)| {
                     let region_span = tracing::debug_span!("region", %region);
+                    // this throws away a possible error from terminate_all,
+                    // but that is ok.
```
jonhoo: Why is it okay? Wouldn't it be better to propagate that error with `?`?

src/providers/aws.rs
akshayknarayan: Alright, I have written a version that collects all the errors without short-circuiting on failures (so that everything that can be cleaned up, is cleaned up). I've probably used this new `Report`/`eyre` stuff incorrectly (tried to model from the baremetal collecting errors example)...

src/providers/aws.rs
```diff
@@ -1282,6 +1285,11 @@ impl RegionLauncher {
                     continue;
                 } else {
                     tracing::warn!("failed to terminate tsunami instances: {}", e);
+                    let e = Err(e).context("failed to terminate tsunami instances");
```
jonhoo: If we fail to terminate the instances, I think we can return early with `?` — the operations below will almost certainly fail.

akshayknarayan: The key stuff might still work?

src/providers/aws.rs
```diff
@@ -1290,23 +1298,34 @@ impl RegionLauncher {
...
+                    Err(e).context("failed to clean up temporary SSH key")?;
+                    unreachable!();
                 }
+
+                Ok(())
             }
             .instrument(key_span)
             .await;
+
+            if r.is_err() {
```
jonhoo: Here too I think we can just early-return on error with `?`. Which should avoid the need for `res`.

src/providers/aws.rs
akshayknarayan: Imagine an example where the spot instance api is not working/failed/whatever, but a security group and key pair were still created. This way we clean those up.

src/providers/aws.rs
```diff
@@ -1327,22 +1351,35 @@ impl RegionLauncher {
                                 tracing::trace!("instances not yet shut down -- retrying");
                                 tokio::time::delay_for(tokio::time::Duration::from_secs(5)).await;
                             } else {
-                                tracing::warn!("failed: {}", err);
-                                break;
+                                Err(Report::msg(err.to_string())
```
```suggestion
                                Err(eyre!(RusotoError::Unkown(r))
```

jonhoo: That way we keep the full underlying error, rather than just its message.

src/providers/aws.rs
```diff
@@ -1315,6 +1334,11 @@ impl RegionLauncher {
                         tracing::warn!(
                             "failed to clean up temporary security group after 60 seconds."
                         );
+
+                        Err(Report::msg(
+                            "failed to clean up temporary security group after 60 seconds.",
+                        ))?;
```
jonhoo, line 1334: Since we return early here, we probably don't also need the `warn!`.

src/providers/aws.rs
```diff
@@ -1327,22 +1351,35 @@ impl RegionLauncher {
...
+                                Err(Report::msg(err.to_string())
+                                    .wrap_err("failed to clean up temporary security group"))?;
+                                unreachable!();
                             }
                         }
                         Err(e) => {
                             tracing::warn!("failed: {}", e);
-                            break;
+                            Err(Report::new(e)
+                                .wrap_err("failed to clean up temporary security group"))?;
```
jonhoo, line 1357: Since we return early here, we probably don't also need the `warn!`.

src/providers/aws.rs
akshayknarayan: From https://docs.rs/eyre/0.4.2/eyre/macro.eyre.html?search= it seems like `eyre!` just uses its argument as a `Debug + Display` anyway? Is there anything extra actually being kept?

akshayknarayan: Doing this results in the following error:
```
error[E0277]: the trait bound `eyre::Report: std::error::Error` is not satisfied
    --> src/providers/aws.rs:1360:94
     |
1360 | ...                   .wrap_err("failed to clean up temporary security group"))?;
     |                                                                                ^ the trait `std::error::Error` is not implemented for `eyre::Report`
     |
     = note: required because of the requirements on the impl of `std::convert::From<eyre::Report>` for `eyre::Report<color_eyre::Context>`
     = note: required by `std::convert::From::from`
```

src/providers/aws.rs
jonhoo: Except those cannot be deleted if the instances haven't been terminated, due to the dependency between them. Basically, they'll both fail with the dependency violation error.

src/providers/aws.rs
jonhoo: @yaahc this seems like an odd error. any idea what causes it?

src/providers/aws.rs
```diff
@@ -554,14 +554,19 @@ where
...
+                    futures_util::future::join_all(self.regions.drain().map(|(region, mut rl)| {
+                        let region_span = tracing::debug_span!("region", %region);
+                        async move { rl.terminate_all().await }.instrument(region_span)
+                    }))
+                    .await;
+                res.into_iter().fold(Ok(()), |acc, x| match acc {
+                    Ok(_) => x,
+                    Err(a) => match x {
+                        Ok(_) => Err(a),
+                        Err(e) => Err(a.wrap_err(e)),
```
jonhoo, line 567: Yeah, this is pretty sad :'(
It's a same instance of the "list of errors" case I discussed with @yaahc in #32.
I think it's okay for now though.

src/providers/aws.rs
akshayknarayan: I don't think that is true for key pairs - they can definitely be deleted even if an instance using that key pair is still active. Also, deleting the key pair only deletes Amazon's copy of the public key. The instance has its own separate copy of the public key. So key pair deletion is independent of instance termination.

jonhoo: Ah, interesting... How about we move up key deletion then? That way, we can then fail here if terminate fails, and the key is deleted regardless.

akshayknarayan: Ah, but what then happens if key pair deletion fails? It's still possible that deleting the instances will work in that case, so that shouldn't short-circuit either?

jonhoo: Yes, I think we move key deletion up, and if it errors, then we just log and continue. If termination errors, then we return an error.

src/providers/aws.rs
jonhoo: It seems to be that the wrapped error does not implement `std::error::Error`, though the error message is then pretty misleading :'(

jonhoo: Although it _looks_ like it does: https://docs.rs/rusoto_core/0.43.0/rusoto_core/enum.RusotoError.html#impl-Error + https://docs.rs/rusoto_ec2/0.43.0/rusoto_ec2/enum.TerminateInstancesError.html#impl-Error.

src/providers/aws.rs
```diff
@@ -1289,40 +1308,14 @@ impl RegionLauncher {
...
-                req.key_name = self.ssh_key_name.clone();
-                if let Err(e) = client.delete_key_pair(req).await {
-                    tracing::warn!("failed to clean up temporary SSH key: {}", e);
-                    Err(e).context("failed to clean up temporary SSH key")?;
+                    // Why is `?` here ok? either:
+                    // 1. there was no spot capacity. So self.instances will be empty, and this
+                    //    block will get skipped, so sg will get cleaned up below.
+                    // 2. there were instances, but we couldn't terminate them. Then, the sg will
+                    //    still be attached to them, so there's no point trying to delete it.
+                    Err(e).context("failed to terminate tsunami instances")?;
```
jonhoo: I think `eyre` favors `wrap_err` over `context` (though I _think_ they both do the same thing).

src/providers/aws.rs
```diff
@@ -1371,15 +1364,10 @@ impl RegionLauncher {
             .instrument(group_span)
             .await;
 
-            if r.is_err() {
-                res = match res {
-                    Ok(_) => r,
-                    Err(e) => Err(e.wrap_err(r.unwrap_err())),
-                };
-            }
+            r?;
```
jonhoo: Can this just go after the `await` above? And the `let r` can go away.

src/providers/aws.rs
```diff
@@ -1281,43 +1308,64 @@ impl RegionLauncher {
...
+                    req.group_id = Some(self.security_group_id.clone());
+                    match client.delete_security_group(req).await {
+                        Ok(_) => break,
+                        Err(RusotoError::Unknown(r)) => {
+                            let err = r.body_as_str();
+                            if err.contains("<Code>DependencyViolation</Code>") {
+                                tracing::trace!("instances not yet shut down -- retrying");
+                                tokio::time::delay_for(tokio::time::Duration::from_secs(5)).await;
+                            } else {
+                                Err(eyre!(err.to_string()))
```
jonhoo: To continue from https://github.com/jonhoo/tsunami/pull/33#discussion_r429313194, does this work:
```rust
Err(Report::new(RusotoError::Unknown(r)))
```

akshayknarayan: @jonhoo see 8a71338

jonhoo (approved):

PR #35, opened 2020-06-04
Improve spot requests cancellation upon timeout

vitorenesduarte:
This PR tries to make sure that cancellation of spot requests always happens when the `max_wait` timeout (passed to `spawn`) triggers. After cancellation, instances are terminated.

Most of the work was moving existing to code to new functions `describe_spot_instance_requests`, `cancel_spot_instance_requests` and `terminate_instances`,  so that these could be called in multiple places.

The "only change" was replacing the `remove` in `self.outstanding_spot_request_ids.remove(id).unwrap()` with a `get`. From what I could understand, it's possible to have more than one spot instance associated with the same spot instance request `id`, in which case the second `remove` would panic. Please let me know if this doesn't make sense. (I've also renamed `outstanding_spot_request_ids` to reflect this change.)

I've tested this a few times with different timeouts to make sure that cancellation is triggered in both `wait_for_spot_instance_requests` (e.g. 1s timeout) and `wait_for_instances` (e.g. 5s timeout).

(I've also made some cosmetic changes while I was getting familiar with the code; I'll revert these, if needed.)

### DIFF
--- a/src/providers/aws.rs
+++ b/src/providers/aws.rs
@@ -970,7 +970,7 @@ impl RegionLauncher {
         &mut self,
         max_wait: Option<time::Duration>,
     ) -> Result<(), Report> {
-        tracing::info!("waiting for {} instances to spawn", self.region.name());
+        tracing::info!("waiting for instances to spawn");
 
         let start = time::Instant::now();
 
@@ -1217,7 +1217,7 @@ impl RegionLauncher {
 
         // terminate instances
         if !self.instances.is_empty() {
-            tracing::info!("terminating {} instances", self.region.name());
+            tracing::info!("terminating instances");
             let instance_ids = self.instances.keys().cloned().collect();
             self.instances.clear();
             // Why is `?` here ok? either:
@@ -1322,10 +1322,7 @@ impl RegionLauncher {
 
     #[instrument(level = "debug")]
     async fn cancel_spot_instance_requests(&self) -> Result<(), Report> {
-        tracing::warn!(
-            "wait time exceeded for {} -- cancelling run",
-            self.region.name()
-        );
+        tracing::warn!("wait time exceeded for -- cancelling run");
         let request_ids = self.spot_requests.keys().cloned().collect();
         let mut cancel = rusoto_ec2::CancelSpotInstanceRequestsRequest::default();
         cancel.spot_instance_request_ids = request_ids;
--- a/src/providers/aws.rs
+++ b/src/providers/aws.rs
@@ -1237,9 +1237,9 @@ impl RegionLauncher {
                 // clean up security groups and keys
                 let start = tokio::time::Instant::now();
                 loop {
-                    if start.elapsed() > tokio::time::Duration::from_secs(2 * 60) {
+                    if start.elapsed() > tokio::time::Duration::from_secs(5 * 60) {
                         Err(Report::msg(
-                            "failed to clean up temporary security group after 120 seconds.",
+                            "failed to clean up temporary security group after 5 minutes.",
                         ))?;
                         unreachable!();
                     }
@@ -1293,6 +1293,9 @@ impl RegionLauncher {
                 let msg = e.to_string();
                 if msg.contains("The spot instance request ID") && msg.contains("does not exist") {
                     tracing::trace!("spot instance requests not yet ready");
+
+                    // let's not hammer the API
+                    tokio::time::delay_for(time::Duration::from_secs(1)).await;
                     continue;
                 } else {
                     res.wrap_err("failed to describe spot instances")?;
--- a/src/providers/aws.rs
+++ b/src/providers/aws.rs
@@ -970,7 +970,7 @@ impl RegionLauncher {
         &mut self,
         max_wait: Option<time::Duration>,
     ) -> Result<(), Report> {
-        tracing::info!("waiting for instances to spawn");
+        tracing::info!("waiting for {} instances to spawn", self.region.name());
 
         let start = time::Instant::now();
         let request_ids = self.spot_requests.keys().cloned().collect();
@@ -1219,7 +1219,7 @@ impl RegionLauncher {
 
         // terminate instances
         if !self.instances.is_empty() {
-            tracing::info!("terminating instances");
+            tracing::info!("terminating {} instances", self.region.name());
             let instance_ids = self.instances.keys().cloned().collect();
             self.instances.clear();
             self.terminate_instances(instance_ids).await?;
@@ -1320,7 +1320,10 @@ impl RegionLauncher {
 
     #[instrument(level = "debug")]
     async fn cancel_spot_instance_requests(&self, request_ids: &Vec<String>) -> Result<(), Report> {
-        tracing::warn!("wait time exceeded -- cancelling run");
+        tracing::warn!(
+            "wait time exceeded for {} -- cancelling run",
+            self.region.name()
+        );
         let mut cancel = rusoto_ec2::CancelSpotInstanceRequestsRequest::default();
         cancel.spot_instance_request_ids = request_ids.clone();
         self.client
--- a/src/providers/aws.rs
+++ b/src/providers/aws.rs
@@ -613,7 +613,7 @@ pub struct RegionLauncher {
     private_key_path: Option<tempfile::NamedTempFile>,
     #[educe(Debug(ignore))]
     client: Option<rusoto_ec2::Ec2Client>,
-    outstanding_spot_request_ids: HashMap<String, TaggedSetup>,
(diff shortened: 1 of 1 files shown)


vitorenesduarte:
I was now wondering what would/should happen if we issue two spot requests targeting different regions, and one triggers cancellation while the other doesn't.

Would something like the following suffice?
```rust
let mut l: tsunami::providers::aws::Launcher<_> = Default::default();
if let Err(e) = l.spawn(ms, Some(std::time::Duration::from_secs(30))).await {
    l.terminate_all().await?;
    return Err(e);
}
```

vitorenesduarte:
It seems that the above indeed makes sure that all instances are terminated but always gives this error "failed to clean up temporary security group after 120 seconds."

jonhoo:
Nice! Let me make a pass through and then I'll get to your comments :)

src/providers/aws.rs
```diff
@@ -969,130 +970,48 @@ impl RegionLauncher {
...
-                                    .unwrap(),
-                            ))
-                        } else {
-                            tracing::error!(%state, "spot request failed: {:?}", &sir.status);
-                            None
-                        }
+                    .map(|(request_id, state, instance_id)| {
+                        assert_eq!(state, "active");
+                        let instance_id = instance_id.unwrap();
+                        let setup = self.spot_requests.get(&request_id).cloned().unwrap();
```
```suggestion
                        let setup = self.spot_requests[&request_id].clone();
```

src/providers/aws.rs
```diff
@@ -1369,6 +1275,120 @@ impl RegionLauncher {
 
         Ok(())
     }
+
+    #[instrument(level = "debug")]
+    async fn describe_spot_instance_requests(
+        &self,
+        request_ids: &Vec<String>,
```
```suggestion
        request_ids: &[String],
```

src/providers/aws.rs
```diff
@@ -969,130 +970,48 @@ impl RegionLauncher {
...
+                    .map(|(request_id, state, instance_id)| {
+                        assert_eq!(state, "active");
+                        let instance_id = instance_id.unwrap();
+                        let setup = self.spot_requests.get(&request_id).cloned().unwrap();
+                        (instance_id, setup)
                     })
                     .collect();
                 break;
-            } else {
-                tokio::time::delay_for(time::Duration::from_secs(1)).await;
```
jonhoo, line 1046: This rate limiting appears to have disappeared?

src/providers/aws.rs
```diff
@@ -969,130 +970,48 @@ impl RegionLauncher {
...
-                            );
-                        }
-                    }
-                }
+                self.cancel_spot_instance_requests(&request_ids).await?;
                 eyre::bail!("wait limit reached");
             }
+
+            // let's not hammer the API
+            tokio::time::delay_for(time::Duration::from_secs(1)).await;
```
jonhoo: This will now not be hit if `wait_limit.is_some()` and `start.elapsed() <= wait_limit`, since that `continue`s early.

src/providers/aws.rs
```diff
@@ -1177,10 +1096,13 @@ impl RegionLauncher {
                 }
             }
 
-            if let Some(to) = max_wait {
-                if time::Instant::now().duration_since(start) > to {
-                    eyre::bail!("timed out");
+            if let Some(wait_limit) = max_wait {
+                if start.elapsed() <= wait_limit {
+                    continue;
```
jonhoo, line 1103: This also now misses the "let's not hammer the API" wait.

src/providers/aws.rs
```diff
@@ -969,130 +970,48 @@ impl RegionLauncher {
         &mut self,
         max_wait: Option<time::Duration>,
     ) -> Result<(), Report> {
-        tracing::info!("waiting for instances to spawn");
+        tracing::info!("waiting for {} instances to spawn", self.region.name());
```
jonhoo: The region name should probably be included through the use of a span (which I think we already do). That way you don't have to tack it into every message. Maybe that span should be `info` level rather than `debug` so that it always gets included?

src/providers/aws.rs
```diff
@@ -1369,6 +1275,120 @@ impl RegionLauncher {
...
+        let client = self.client.as_ref().unwrap();
+        let mut termination_req = rusoto_ec2::TerminateInstancesRequest::default();
+        termination_req.instance_ids = instance_ids;
+        while let Err(e) = client.terminate_instances(termination_req.clone()).await {
+            let msg = e.to_string();
+            if msg.contains("Pooled stream disconnected") || msg.contains("broken pipe") {
+                tracing::trace!("retrying instance termination");
+                continue;
+            } else {
+                // Why is `?` here ok? either:
```
jonhoo: This comment should be on the `?` where `terminate_instances` is awaited, not here inside `terminate_instances`.

jonhoo:
This looks pretty good! And thanks for also doing some cleanup along the way :)

As for your questions:
> From what I could understand, it's possible to have more than one spot instance associated with the same spot instance request id, in which case the second remove would panic.

My experience is actually that this doesn't happen. Even if you tell AWS that you want multiple instances for a single request, it gives them _different_ spot request ids. But it seems fine to future-proof against that anyway.

> I was now wondering what would/should happen if we issue two spot requests targeting different regions, and one triggers cancellation while the other doesn't.

I actually don't know what will happen today. I think "ideally" what should happen is that everything that was launches as part of _that_ `spawn_all` should be torn down, but that may be tricky to achieve in practice.

> It seems that the above indeed makes sure that all instances are terminated but always gives this error "failed to clean up temporary security group after 120 seconds."

Oh, weird. Not sure why that is. If you increase the time limit, does that make it work?

jonhoo:
@akshayknarayan Would you also mind taking a quick look, since you also recently made some big edits to this code? It looks fine to me as-is.

src/providers/aws.rs
vitorenesduarte: I've tried this but making that span `info` doesn't seem to be enough. I think that the problem is in the instrument `level` of `launch` and e.g. `wait_for_spot_instance_requests` (`debug` and `trace`, respectively), since I can see the region in the logs after changing them to `info`. (I've reverted the change).

src/providers/aws.rs
```diff
@@ -1369,6 +1278,112 @@ impl RegionLauncher {
...
+        let request_ids = self.spot_requests.keys().cloned().collect();
+        let mut req = rusoto_ec2::DescribeSpotInstanceRequestsRequest::default();
+        req.spot_instance_request_ids = Some(request_ids);
+        loop {
+            let res = client.describe_spot_instance_requests(req.clone()).await;
+            if let Err(ref e) = res {
+                let msg = e.to_string();
+                if msg.contains("The spot instance request ID") && msg.contains("does not exist") {
+                    tracing::trace!("spot instance requests not yet ready");
+                    continue;
```
akshayknarayan, line 1299: Will this hammer the api? Should we add a delay?

akshayknarayan (commented):
One question: there is now a nested-`loop{}` for `wait_for_spot_instance_requests -> describe_spot_instance_requests`. Is this intentional? The flow changes from checking all the spot requests at once, to checking one at a time?

vitorenesduarte:
> Oh, weird. Not sure why that is. If you increase the time limit, does that make it work?

Yes, changed it to 5 minutes and it worked. Thanks. The issue was exactly what the code is already prepared for: instances were shutting down. Should I push that change as well?

src/providers/aws.rs
jonhoo: Huh, that's interesting. Maybe they don't propagate spans if their level isn't met. That sounds like a bug, let me check.

src/providers/aws.rs
vitorenesduarte: Yes, thanks. I've added a delay.

vitorenesduarte: > One question: there is now a nested-`loop{}` for `wait_for_spot_instance_requests -> describe_spot_instance_requests`. Is this intentional? The flow changes from checking all the spot requests at once, to checking one at a time?

I believe the flow didn't change. However, this loop doesn't have a timeout. From my limited experience it's not a problem in practice, but maybe this loop should timeout as well?

akshayknarayan: Ah, I see now :)
I think the only case where the inner loop would loop forever is if the spot instance request ID never showed up? I agree it's probably safe to assume the aws api will report its existence eventually, since `make_spot_instance_requests` succeeded at this point.

vitorenesduarte: That makes sense. I'll leave it as is then.

jonhoo (approved):

jonhoo:
Great, thank you!

vitorenesduarte:
Awesome! Thanks for the quick review.

PR #36, opened 2020-06-04
Bump rusoto

vitorenesduarte:
0.44.0 was released and has support for two new regions: `AfSouth1` and `EuSouth1`.


jonhoo (approved):

jonhoo:
This is a backwards-incompatible change, but luckily we're in alpha, so that's fine!

Issue #37, opened 2020-06-04
Increase ssh session timeout


vitorenesduarte:
I have something similar to the ping example where every region is continuously pinging every other region for 30 minutes (with `ping -w 1800`). This experiment failed with the error `the connection was terminated` and I can see this string in `./target/debug/deps/libopenssh-9b9063468e4fbdfc.rlib`.

Internet suggests that this was a timeout in ssh which could be prevented by setting the `ServerAliveInterval` ssh option to 0 (it would never timeout in this case). I was going to open a PR in `openssh` to allow configuring  this option (and maybe also`ServerAliveCountMax`) in [`openssh::SessionBuilder`](https://docs.rs/openssh/0.6.1/openssh/struct.SessionBuilder.html) but I believe this is not enough since the session is created internally in `tsunami`. Given that, I'm not sure how to proceed. Do you have ideas? Thanks.

jonhoo:
Hmm, it's a good question. I don't think that's quite how `ServerAliveInterval` works though? As I understand it, `ServerAliveInterval` is how often the client should send keep-alive messages to the server. Setting that to 0 means "never send keep-alives", which doesn't seem like what you want. I believe it is the _server_ that terminates the ssh connection if it is idle for too long, so what you actually want is to set `ServerAliveInterval` to some **non**-zero number to indicate that a keep-alive _should_ be sent.

You could set this in your `.ssh/config`:
```
Host *
    ServerAliveInterval 240
```
And openssh should pick that up just fine. Having it as an option on `openssh::SessionBuilder` seems pretty reasonable too though.

As for how to pass that through to `tsunami`, maybe one thing to do would be to have a version of [`connect_all`](https://docs.rs/tsunami/0.11.0-beta.4/tsunami/trait.Tsunami.html#tymethod.connect_all) that takes a closure that's given a `&mut openssh::SessionBuilder`, and can modify it to set any additional parameters?

vitorenesduarte:
That makes sense, thanks. I'll try to come up with something.

vitorenesduarte:
It turns out that the error `the connection was terminated` is not related with the `ServerAliveInterval`. In fact, I believe this value was already non-zero due to `BatchMode=yes`, which sets it to 300.

[...]

PR #38, opened 2020-06-04
WIP: Add Setup::ssh_setup

vitorenesduarte:
As opposed to what was discussed in #37, I've added an `ssh_setup` method to each provider's `Setup`. The closure passed to this method is then forwarded to the `MachineDescriptor::connect_ssh` where the `SessionBuilder` is created.
(Initially I tried to add a method similar to `connect_all` but noticed that ssh connections are created before when setting up the machine).

Could something like this work? Thanks.
(I still need to test it and add documentation.)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -148,6 +148,7 @@ pub use ssh::Session;
 use std::collections::HashMap;
 use std::future::Future;
 use std::pin::Pin;
+use std::sync::Arc;
 use tracing::instrument;
 
 pub mod providers;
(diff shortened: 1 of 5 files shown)


jonhoo:
Hmm, I feel like the setup mutations probably aren't needed for the early connections, and fresh ones are (I believe) constructed when you call `connect_all` anyway. But probably doesn't hurt to have them all use it.

jonhoo:
Future self, see https://github.com/jonhoo/tsunami/issues/37#issuecomment-639532289 for why this was closed.

PR #39, opened 2020-06-05
WIP: Make private key file not temporary

vitorenesduarte:
For debugging reasons, sometimes it may be useful to ssh into the spawned spot instances. Currently, I believe this is not possible/easy. I was thinking of storing the private key in the current directory and name it something like `region_name.key`. Could something like this work?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ maintenance = { status = "passively-maintained" }
 
 [features]
 default = ["aws", "azure", "baremetal"]
-aws = ["rusoto_core", "rusoto_ec2", "futures-util", "tempfile", "ubuntu-ami"]
+aws = ["rusoto_core", "rusoto_ec2", "futures-util", "ubuntu-ami"]
 azure = ["serde", "serde_json", "futures-util", "tokio", "tokio/process"]
 baremetal = []
 args = ["structopt"]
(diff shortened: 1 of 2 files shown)


jonhoo:
This is already possible by enabling debug logging, and then just looking at what keyfile path is used for which instance. We could change the directory that holds those keys to be in the current directory instead of the normal temporary file path (commonly `/tmp`) if that would help, but I still feel like they should be temporary files that are dropped when the machines are terminated.

vitorenesduarte:
Ah, I see. One issue is that debugging seems to only be possible while the program is running, since the keys are deleted on `Drop` (https://docs.rs/tempfile/3.1.0/tempfile/struct.NamedTempFile.html#resource-leaking), not when machines are terminated. But knowing this, I'll come up with something that works. Thanks and sorry for the noise.

jonhoo:
No worries, not noise at all! One potentially good change would be to make sure the key files don't get dropped until the machines go away.

vitorenesduarte:
> One potentially good change would be to make sure the key files don't get dropped until the machines go away.

@jonhoo would something like this be acceptable? https://github.com/jonhoo/tsunami/compare/master...vitorenesduarte:key_pair?expand=1

- adding `private_key` to `Machine` allows users to make the key persistent, if needed; together with `username`, this makes `tsunami` more flexible since users have now the necessary information to use something different than `openssh::Session`
- there's also a breaking change motivated by the last point: `setup` now takes `Machine` instead of `openssh::Session`

jonhoo:
Hmm, interesting. I think I like that change, though I think it should be a `&Machine`, not a `&mut Machine`, since we don't want the user modifying the entries of `Machine`. This also works out nicely because you generally only ever need `&openssh::Session`, not `&mut`.

PR #41, opened 2020-06-22
Make openssh optional

vitorenesduarte:


vitorenesduarte:
Still needs some cleanup, but if it looks good, I'll update it and write some documentation.

Since `openssh` is optional now, the `aws` feature also needs `tokio`. With that, the following works:

```bash
for provider in aws azure baremetal; do
  cargo check --no-default-features --features ${provider}
  cargo check --no-default-features --features "${provider},openssh"
done
```

src/lib.rs
```diff
@@ -181,8 +179,13 @@ pub struct Machine<'tsunami> {
     /// equivalent to `public_ip`.
     pub public_dns: String,
 
+    #[cfg(feature = "openssh")]
```
jonhoo, line 226: This won't work since cargo features have to be additive. You'll want to add `#[non_exhausive]` to the struct, then we should be fine.

jonhoo:
You'll probably want to add something like this to `azure-pipelines.yml`:
https://github.com/jonhoo/inferno/blob/bcd8fc295ec6d74c03e516388efd7c26a3307c8c/azure-pipelines.yml#L11-L22

jonhoo:
That's a neat change! I don't know how often people will _not_ want an SSH connection, but fair enough :)

src/lib.rs
vitorenesduarte: Could you explain why this is non-additive? I've also read a bit about `#[non_exhaustive]` but couldn't understand how it plays with cargo features.

jonhoo: Ah, actually, in this particular case the private `_tsunami` field saves us, because it essentially emulates `#[non_exhausive]`. But I still think it'd be good to add. The basic idea is that a user _could_ have written this code (if all the fields were public):
```rust
let Machine {
  nickname,
  public_ip,
  private_ip,
  public_dns,
  username,
  private_key,
} = vm;
```
In a crate, call it `A`, that depends on `tsunami` _without_ the `openssh` feature. Now imagine some other crate, `B`, which also depends on `tsunami`, but _with_ the `openssh` feature. In isolation, they both compile just fine. But now imagine some crate `C` that depends on both `A` and `B`. Cargo only builds one instances of every crate (per major version), so `tsunami` will only be built once. And since `B` requires the `openssh` feature, `tsunami` will be built with that feature enabled (this is the additive part — Cargo takes the _union_ of all the feature requirements). But this will make `A` not compile — the compiler will complain (correctly) that the pattern match above is missing the field `ssh`. The same problem applies to crates that construct a `Machine` by explicitly listing all its fields.

`#[non_exhausive]` makes it so that a type cannot be constructed using its implicit constructor (i.e., by listing its fields), and so that it cannot be matched against without including the wildcard pattern (`..` for fields, `_` for enum variants).

vitorenesduarte: That makes total sense now. Thanks for the awesome reply!

vitorenesduarte:
> You'll probably want to add something like this to `azure-pipelines.yml`:
> https://github.com/jonhoo/inferno/blob/bcd8fc295ec6d74c03e516388efd7c26a3307c8c/azure-pipelines.yml#L11-L22

Thanks, didn't know about `cargo-hack`. I believe I spotted a typo :p https://github.com/jonhoo/inferno/pull/184

src/lib.rs
jonhoo: I'm glad! The other thing we have to be careful with here is that the added fields cannot change any aspect of the containing type. For example, it would not be okay if including `openssh::Session` made the type `!Sync` or `!Send` (I don't believe that's the case here), because otherwise you end up in non-additive land again. Specifically, imagine that `A` depends on `Machine` being `Send`, but it stops being `Send` once `B` adds the `openssh` feature. `A` would then stop compiling.

jonhoo (approved):

jonhoo:
@akshayknarayan and @znewman01 I'd love your input on this too.

src/providers/mod.rs
```diff
@@ -192,7 +192,7 @@ async fn setup_machine(
     let mut m = m.connect_ssh(username, private_key, max_wait, 22).await?;
 
     tracing::debug!("setting up instance");
-    f(&mut m.ssh).await.wrap_err("setup procedure failed")?;
+    f(&mut m).await.wrap_err("setup procedure failed")?;
```
akshayknarayan, line 195: The setup function won't work if there is no ssh, so the existence of the setup functions should probably go away if the openssh feature is disabled.

akshayknarayan (commented):
I guess the intended use of this is a program which just launches a bunch of machines and dumps their connection info to the screen, and then the user connects separately?

src/lib.rs
```diff
@@ -198,7 +246,33 @@ impl<'t> MachineDescriptor<'t> {
...
+            _tsunami: self._tsunami,
+            #[cfg(feature = "openssh")]
+            ssh: sess,
+            username: username.to_string(),
+            private_key: key_path.map(|path| path.to_path_buf()),
+        })
+    }
+
+    #[cfg(feature = "openssh")]
+    async fn do_connect_ssh(
```
znewman01: [nit] I understand the desire to minimize the surface area of the change, but if `connect_ssh` in some cases *doesn't* connect to SSH it's probably no longer a useful name.

Maybe `connect` or `initialize` and then you can have a private `connect_ssh` (what you call `do_connect_ssh` now)?

Cargo.toml
```diff
@@ -17,8 +17,8 @@ azure-devops = { project = "jonhoo/jonhoo", pipeline = "tsunami", build = "25" }
 maintenance = { status = "passively-maintained" }
 
 [features]
-default = ["aws", "azure", "baremetal"]
-aws = ["rusoto_core", "rusoto_ec2", "futures-util", "tempfile", "ubuntu-ami"]
+default = ["aws", "azure", "baremetal", "openssh"]
```
znewman01, line 20: Worth sticking a blurb about this in the crate documentation?

In particular, you could give an example or two of when it's useful :)

znewman01: Also `Tsunami::connect_all` maybe deserves a note

znewman01 (commented):
For the most part looks good! Some nits.

I suppose for completeness you could also get rid of generating a keypair etc. in each of the providers but probably not strictly necessary

vitorenesduarte:
> I guess the intended use of this is a program which just launches a bunch of machines and dumps their connection info to the screen, and then the user connects separately?

It allows you to connect using something other than `openssh`. I've started this because I couldn't tell why my commands were not working (maybe some escaping issue), and moved to using `tokio::process::Command` instead, which was easier to debug. I can't say if this is motivation enough, though.

src/providers/mod.rs
vitorenesduarte: I believe it does work. A setup function that uses `Machine.ssh` won't compile unless the `openssh` feature is enabled.

Cargo.toml
vitorenesduarte: > Maybe connect or initialize and then you can have a private connect_ssh (what you call do_connect_ssh now)?

I've renamed `connect_ssh` to `connect`.

> Worth sticking a blurb about this in the crate documentation?
> 
> In particular, you could give an example or two of when it's useful :)

Could you say something more about this? I've already added some documentation about these cargo features.

> Also `Tsunami::connect_all` maybe deserves a note

Good point. I believe the only reason to call `connect_all` without `openssh` is to get to `Machine`s.

src/lib.rs
```diff
@@ -191,14 +239,38 @@ pub struct Machine<'tsunami> {
 impl<'t> MachineDescriptor<'t> {
     #[cfg(any(feature = "aws", feature = "azure", feature = "baremetal"))]
     #[instrument(level = "debug", skip(key_path, timeout))]
-    async fn connect_ssh(
+    async fn connect(
```
jonhoo, line 242: This method is just all-around pretty weird when `openssh` is disabled, since it doesn't connect at all. I wonder if it simply shouldn't be available without the `openssh` feature? Without it, users can just use `MachineDescriptor` directly instead.

jonhoo:
As an aside — best to not force-push while people are reviewing. It makes threads really hard to follow (mostly because GitHub doesn't handle force pushes very well).

vitorenesduarte:
> I suppose for completeness you could also get rid of generating a keypair etc. in each of the providers but probably not strictly necessary

The idea of the PR is to also expose the private key so that it can be used to SSH into the machine, either manually (e.g. for debugging), or with something different from `openssh`. I apologize, I should have written this motivation when opening the PR.

src/providers/mod.rs
akshayknarayan: You are right, I meant to say that it won't be useful, but as you say it is possible to use a different connection method.

src/lib.rs
vitorenesduarte: `MachineDescriptor` is not public, only `Machine`. This method does two things: maybe connects (if `openssh` is enabled) and builds the `Machine` to be later returned to the user. Maybe rename it instead to `initialize` as suggested by @znewman01?

jonhoo: I guess I'm wonder if, with this feature, `MachineDescriptor` _should_ be public, since that's basically all we expose if someone doesn't want to use `openssh`. Not entirely sure what that would look like though...

vitorenesduarte: Hmm, making `openssh` optional may be indeed a bad idea. I was trying to figure out how to get rid of `MachineDescriptor` and found out that (at least) in the AWS provider, openssh is used to check if instances are ready: https://github.com/jonhoo/tsunami/blob/master/src/providers/aws.rs#L1076-L1095

vitorenesduarte:
Closing this in favor of #42.

PR #43, opened 2020-06-26
Set Machine's public_dns in AWS and private_ip in Azure

vitorenesduarte:
This PR sets `Machine.public_dns` in AWS and `Machine.private_ip` in Azure (information we had, that was just not being recorded).

While doing this change, the logic of "setting public_dns equal to public_ip if there's no public_dns" was moved up to `connect_ssh`. Now this rule lives in this single place.

(For some sanity, I've also reordered some things to make sure that the following order is maintained: `public_dns`, `public_ip`, and `private_ip`)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -252,14 +252,10 @@ impl<'t> MachineDescriptor<'t> {
         tracing::trace!("connected");
 
         let public_ip = self.public_ip;
-        // if not defined, set public dns to be the public ip
-        let public_dns = self
-            .public_dns
-            .map(String::from)
-            .unwrap_or_else(|| public_ip.clone());
         Ok(Machine {
             nickname: self.nickname,
-            public_dns,
+            // if not defined, set public dns to be the public ip
+            public_dns: self.public_dns.unwrap_or_else(|| public_ip.clone()),
             public_ip,
             private_ip: self.private_ip,
             _tsunami: self._tsunami,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -186,9 +186,9 @@ pub mod providers;
 #[derive(Debug)]
 struct MachineDescriptor<'tsunami> {
     pub(crate) nickname: String,
+    pub(crate) public_dns: Option<String>,
     pub(crate) public_ip: String,
     pub(crate) private_ip: Option<String>,
-    pub(crate) public_dns: String,
 
     // tie the lifetime of the machine to the Tsunami.
     _tsunami: std::marker::PhantomData<&'tsunami ()>,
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -253,9 +253,13 @@ impl<'t> MachineDescriptor<'t> {
 
         Ok(Machine {
             nickname: self.nickname,
+            // if not defined, set public dns to be the public ip
+            public_dns: self
+                .public_dns
+                .map(String::from)
+                .unwrap_or(self.public_ip.to_string()),
```
jonhoo: This should technically be an `unwrap_or_else(||`, otherwise you'll end up allocating an extra string on each invocation.

jonhoo (approved):

src/lib.rs
vitorenesduarte: I've tried that but I'm getting the following error:
```
error[E0382]: borrow of moved value: `self`
   --> src/lib.rs:260:33
    |
257 |               public_dns: self
    |  _________________________-
258 | |                 .public_dns
    | |___________________________- value moved here
259 |                   .map(String::from)
260 |                   .unwrap_or_else(|| self.public_ip.to_string()),
    |                                   ^^ ---- borrow occurs due to use in closure
    |                                   |
    |                                   value borrowed here after move
    |
    = note: move occurs because `self.public_dns` has type `std::option::Option<std::string::String>`, which does not implement the `Copy` trait
```

I'm not sure how to fix this.

jonhoo: Ah, yes, that's because you're moving out of `self` in the same constructor (when you construct all the other fields). If you just do it separately above, it should be fine:
```rust
let public_dns = self
                .public_dns
                .unwrap_or(self.public_ip.to_string());
```
and then use `public_dns` instead of `self.public_dns` below.

vitorenesduarte: I had to first move `self.public_ip` to its own variable, and only then compute `self.public_dns`:

```rust
        let public_ip = self.public_ip;
        // if not defined, set public dns to be the public ip
        let public_dns = self
            .public_dns
            .map(String::from)
            .unwrap_or_else(|| public_ip.clone());
```

I guess there was no need to also `public_dns` outside of `Machine`'s constructor.

jonhoo: Ah, yes, that's because Rust only sees the closure borrow _all of `self`_, not just `self.public_ip`. This is because the compiler's understanding of split/partial borrows is limited to a single scope (it can't, for example, see "through" a closure for what fields it is accessing).

src/lib.rs
```diff
@@ -251,11 +251,17 @@ impl<'t> MachineDescriptor<'t> {
         let sess = sess.connect(&self.public_ip).await?;
         tracing::trace!("connected");
 
+        let public_ip = self.public_ip;
+        // if not defined, set public dns to be the public ip
+        let public_dns = self
+            .public_dns
+            .map(String::from)
```
jonhoo: Why do you need this `map`? Isn't `public_dns` already `Option<String>`?

vitorenesduarte: Ah yes. Bad copy-paste 😅

src/lib.rs
vitorenesduarte: That makes sense, thanks! I've also had this problem with methods that borrow self: even though a single field is accessed, all self is borrowed. I guess it's the same problem. Do you know if there are any plans to address this?

jonhoo: I think the thing to watch is probably https://github.com/rust-lang/rfcs/issues/1215, though not sure if that'll solve this particular pain point :shrug:

vitorenesduarte: Thanks!

PR #45, opened 2020-07-28
Add support for launching on-demand AWS instances

### LINKED ISSUE #25: Add support for on-demand AWS instance
I have a horrible hacky implementation on #23: https://github.com/znewman01/tsunami/commit/8c91e80513f6c587621d2ac30a88e8ed046e2430

Which I can clean up (at some point)

jonhoo:
I've done a small-scale test of it, and it seems to be working!

Fixes #25.

/cc @akshayknarayan @znewman01

### DIFF
--- a/src/providers/aws.rs
+++ b/src/providers/aws.rs
@@ -109,6 +109,38 @@ use std::time;
 use tracing::instrument;
 use tracing_futures::Instrument;
 
+/// Dictate how a set of instances should be launched.
+#[derive(Debug, Clone)]
+#[allow(missing_copy_implementations)]
+#[non_exhaustive]
+pub enum LaunchMode {
+    /// Use AWS [defined duration](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-requests.html#fixed-duration-spot-instances) spot instances.
+    DefinedDuration {
+        /// The lifetime of the defined duration instances.
+        /// This value must be between 1 and 6 hours.
+        hours: usize,
+    },
+    /// Use regular AWS on-demand instances.
+    OnDemand,
+}
+
+impl LaunchMode {
+    /// Launch using AWS [defined duration](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-requests.html#fixed-duration-spot-instances) spot instances.
+    ///
+    /// The lifetime of such instances must be declared in advance (1-6 hours).
+    /// This method thus clamps `hours` to be between 1 and 6.
+    pub fn duration_spot(hours: usize) -> Self {
+        let hours = std::cmp::min(hours, 6);
+        let hours = std::cmp::max(hours, 1);
+        Self::DefinedDuration { hours }
+    }
+
+    /// Launch using regular AWS on-demand instances.
+    pub fn on_demand() -> Self {
+        Self::OnDemand
+    }
+}
+
 /// Available configurations of availability zone specifiers.
 ///
 /// See [the aws docs](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#using-regions-availability-zones-launching) for more information.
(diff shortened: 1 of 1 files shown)


src/providers/aws.rs
```diff
@@ -109,6 +109,38 @@ use std::time;
 use tracing::instrument;
 use tracing_futures::Instrument;
 
+/// Dictate how a set of instances should be launched.
+#[derive(Debug, Clone)]
+#[allow(missing_copy_implementations)]
```
akshayknarayan, line 115: What's wrong with having it be copy?

src/providers/aws.rs
```diff
@@ -827,6 +877,118 @@ impl RegionLauncher {
         Ok(self)
     }
 
+    #[instrument(level = "trace", skip(self))]
+    async fn make_on_demand_requests<M>(&mut self, machines: M) -> Result<(), Report>
```
akshayknarayan: I wonder if it might be possible to share more of this code with the other implementation, but this seems fine.

akshayknarayan: Maybe a few utility functions to e.g. do the instance grouping and make placement groups?

akshayknarayan (commented):
Seems good! We probably want to also update this line of the module docs: https://github.com/jonhoo/tsunami/pull/45/files#diff-60699702750aef5a8cdc050b4fce64eaL7

src/providers/aws.rs
jonhoo: In general, making types `Copy` is a backwards-compatibility risk. It means that if we _later_ add a field that can't be `Copy` (like a `String`), then we'd have to do a breaking release. In general, you should only implement `Copy` for public types if you think it is highly unlikely that they might gain non-`Copy` members in the future. For this, I think there's a decent chance we may add a `String` field later, so no `Copy` :)

src/providers/aws.rs
jonhoo: Oh, yeah, probably. I'll do a pass and see if I can tidy it up a little.

src/providers/aws.rs
```diff
@@ -109,6 +110,38 @@ use std::time;
...
+    },
+    /// Use regular AWS on-demand instances.
+    OnDemand,
+}
+
+impl LaunchMode {
+    /// Launch using AWS [defined duration](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-requests.html#fixed-duration-spot-instances) spot instances.
+    ///
+    /// The lifetime of such instances must be declared in advance (1-6 hours).
+    /// This method thus clamps `hours` to be between 1 and 6.
```
znewman01, line 132: This is definitely personal preference but I'd rather barf on invalid input than silently succeed. Totally fine either way as long as it's documented

znewman01: Waiting on those [range types](https://github.com/rust-lang/rfcs/issues/671) I guess :)

src/providers/aws.rs
```diff
@@ -827,6 +880,134 @@ impl RegionLauncher {
         Ok(self)
     }
 
+    #[instrument(level = "trace", skip(self, mk))]
+    async fn make_placement<R>(
```
znewman01, line 889: I know this is private but worth documenting briefly since the point of `mk` is a little nonobvious from its type/name.

znewman01 (approved):
handful of purely-aesthetic comments, main logic looks good to me

i guess i have a longer-term question about how to test this library but it's out-of-scope for now :)

src/providers/aws.rs
jonhoo: Yeah, this one was a tricky decision. The main issue I have here is that panicking would mean that cleanup isn't run, which means that _past_ instances may be left running. We could define an `enum` I suppose, but that seems like overkill. My inclination is to keep it as-is.

jonhoo:
Yeah, I also would love to have this library have decent testing. But how to do that properly, I have no idea...

PR #47, opened 2020-10-07
TrySpot Mode

akshayknarayan:
Add the ability to try for spot instances, but fail back to on-demand instances if there was no spot capacity.

### DIFF
--- a/src/providers/aws.rs
+++ b/src/providers/aws.rs
@@ -144,6 +144,19 @@ impl LaunchMode {
         Self::DefinedDuration { hours }
     }
 
+    /// Try to launch using AWS [defined
+    /// duration](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-requests.html#fixed-duration-spot-instances)
+    /// spot instances, and fall back to OnDemand instances otherwise.
+    ///
+    /// The lifetime of such instances must be declared in advance (1-6 hours).
+    /// This method thus clamps `hours` to be between 1 and 6.
+    pub fn try_duration_spot(hours: usize) -> Self {
+        match Self::duration_spot(hours) {
+            Self::DefinedDuration { hours } => Self::TrySpot { hours },
+            _ => unreachable!(),
+        }
+    }
+
     /// Launch using regular AWS on-demand instances.
     pub fn on_demand() -> Self {
         Self::OnDemand
--- a/src/providers/aws.rs
+++ b/src/providers/aws.rs
@@ -115,8 +115,16 @@ use tracing_futures::Instrument;
 #[allow(missing_copy_implementations)]
 #[non_exhaustive]
 pub enum LaunchMode {
-    /// Use AWS [defined duration](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-requests.html#fixed-duration-spot-instances) spot instances.
+    /// Use AWS [defined duration](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-requests.html#fixed-duration-spot-instances)
+    /// spot instances. Fails with an error if there is no spot capacity.
     DefinedDuration {
+        /// The lifetime of the defined duration instances.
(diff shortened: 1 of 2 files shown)


src/providers/aws.rs
```diff
@@ -727,41 +735,69 @@ impl RegionLauncher {
     /// Make spot instance requests, wait for the instances, and then call the
     /// instance setup functions.
     #[instrument(level = "debug", skip(self, max_wait))]
-    pub async fn launch<M>(
-        &mut self,
-        mode: LaunchMode,
+    pub async fn launch<'l, M>(
```
jonhoo: Hmm, why is this lifetime annotation now suddenly needed?

src/providers/aws.rs
```diff
@@ -727,41 +735,69 @@ impl RegionLauncher {
...
-                .wrap_err("failed to make spot instance requests")?;
+        let machines: Vec<_> = machines.into_iter().collect();
 
-                let start = time::Instant::now();
-                self.wait_for_spot_instance_requests(max_wait)
-                    .await
-                    .wrap_err("failed while waiting for spot instances fulfilment")?;
-                if let Some(ref mut d) = max_wait {
-                    *d -= time::Instant::now().duration_since(start);
+        loop {
```
jonhoo: I don't think this needs to be a `loop`. Just check spot first, and early-return if it succeeds?

src/providers/aws.rs
```diff
@@ -727,41 +735,69 @@ impl RegionLauncher {
...
+                        .await
+                        .wrap_err(eyre!(
+                            "failed while waiting for spot instances fulfilment in {}",
+                            self.region.name()
+                        ))
+                    {
+                        if let LaunchMode::TrySpot { .. } = mode {
+                            mode = LaunchMode::OnDemand;
+                            tracing::debug!(err = ?e, "re-trying with OnDemand instace");
+                            continue;
```
jonhoo: This doesn't seem quite right. What if some spot instances started but not others? I don't think any code will actually clean those up with the code as it is now?

src/providers/aws.rs
akshayknarayan: With the loop, we end up not having to copy-paste the call to `wait_for_instances` twice. This also relates to why there was a lifetime hanging around from my first attempt at this - I tried a way with a closure with early return instead of the loop, but ended up deciding the result was too complicated.
Another way is to have a helper function that does this part, if you think that would be more readable.

src/providers/aws.rs
akshayknarayan: Ahh, in my head everything still gets cleaned up on drop.

akshayknarayan: actually, `wait_for_spot_instance_requests` will handle this via `cancel_spot_instance_requets` I think.

src/providers/aws.rs
```diff
@@ -748,23 +763,45 @@ impl RegionLauncher {
...
+                    .wrap_err(eyre!(
+                        "failed while waiting for spot instances fulfilment in {}",
+                        self.region.name()
+                    ))
+                {
+                    // if wait_for_spot_instance_requests returned an Err, it will have cleaned up
+                    // the spot instance requests already.
+                    if let LaunchMode::TrySpot { .. } = mode {
+                        tracing::debug!(err = ?e, "re-trying with OnDemand instace");
+                        do_ondemand = true;
```
jonhoo, line 794: Hmm, this won't work. The `Err(e)?` is still executed, which will return early. I think you need an `else` here?

src/providers/aws.rs
jonhoo: Ah, yes, I think you're right!

src/providers/aws.rs
akshayknarayan: yeah, good catch :+1:

jonhoo (approved):

jonhoo:
Looks good, thank you!

PR #49, opened 2021-03-09
Fix TrySpot launch mode

akshayknarayan:
If the timer expired, we shouldn't be trying to subtract the elapsed time from it.

### DIFF
--- a/src/providers/aws.rs
+++ b/src/providers/aws.rs
@@ -795,10 +795,10 @@ impl RegionLauncher {
                     } else {
                         return Err(e);
                     }
-                }
-
-                if let Some(ref mut d) = max_wait {
-                    *d -= time::Instant::now().duration_since(start);
+                } else {
+                    if let Some(ref mut d) = max_wait {
+                        *d -= time::Instant::now().duration_since(start);
+                    }
                 }
             }
             LaunchMode::OnDemand => {


akshayknarayan:
not really sure what's going on with rustfmt...

jonhoo:
It looks like it disagrees with some of your local formatter's decisions?
https://dev.azure.com/jonhoo/jonhoo/_build/results?buildId=2049&view=logs&jobId=3f85e25b-fb23-5eb7-8385-51966a33df73&j=3f85e25b-fb23-5eb7-8385-51966a33df73&t=76a09b4e-1ffe-5071-647c-4f3a37660459

jonhoo:
Or rather, with the formatting choices of past me :sweat_smile:

jonhoo:
This fix looks a little odd to me. Aren't there other reasons why the call may have failed than just if the timer expired?

akshayknarayan:
There are only two potentially time-consuming reasons the call to `wait_for_spot_instance_requests` might have failed. The first is the timer expiring, and the second is the spot request unexpectedly closing (https://github.com/jonhoo/tsunami/blob/master/src/providers/aws.rs#L1221). 

In the first case we definitely want to not subtract from the timer (even a saturating_sub to fix the overflow), because then any subsequent timed call we make when trying with OnDemand mode will spuriously fail because the max_wait will be 0.

In the second case, we might end up waiting longer than the max_wait because we didn't subtract the timer and `wait_for_spot_instance_requests` took a large fraction of max_wait time to fail, but I think this edge case is not really worth worrying about. If we wanted to fix this, the best way would probably be to have a separate max_wait as part of Mode::TrySpot, and the original max_wait is passed through.

jonhoo (approved):
Seems reasonable!

PR #51, opened 2021-10-14
VPC Support (for AWS)

bsodmike:
This is a quick PR to address #1
I haven't done much testing beyond keeping the rust analyzer happy.

There's improvement for validating the ipv4 CIDR block, which at the moment is just a `String`, and also definitely room for ipv6 (which is currently ignored).

Will give this another look with fresh eyes later on. Opening this to open the discussion, thanks!

### DIFF
--- a/src/providers/aws.rs
+++ b/src/providers/aws.rs
@@ -224,6 +224,7 @@ pub struct Setup {
                 + 'static,
         >,
     >,
+    default_vpc_cidr_block: String,
 }
 
 impl super::MachineSetup for Setup {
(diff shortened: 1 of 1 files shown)


jonhoo:
Hmm, I'm hesitant to land something like this without actual testing :sweat_smile: 
I also think it's fairly likely that something like Terraform is the way to go for use-cases like what tsunami was originally built for these days, and it may not be worth extending it a lot further :p

bsodmike:
> Hmm, I'm hesitant to land something like this without actual testing sweat_smile I also think it's fairly likely that something like Terraform is the way to go for use-cases like what tsunami was originally built for these days, and it may not be worth extending it a lot further :p

No worries - I was looking to contribute to something Rust related; Even use Terraform myself, it's pretty much the standard now.  Will look to help maintain something else; if you have any suggestions let me know! Thanks!

jonhoo:
Yeah, I think I'll probably just archive this project at this point, especially since I don't use it myself any more :sweat_smile: 

Thanks for looking into it though! I think you energy is probably best spent either on inferno or on hdrhistogram :+1:

